Issue #57, opened 2019-09-19
Feature Request


davelab6:
https://devblogs.microsoft.com/commandline/cascadia-code/ says

> 👉 Note: There is currently no proper testing framework built into the repository, so all pull requests will have to be manually tested and validated. This will slow down pull request approvals, so please bear with us. 😊

The GitHub.com/googlefonts/FontBakery font testing tool may be suitable for meeting this need. It has been successfully used via Travis CI to do regression testing.

cc @felipesanches @graphicore @m4rc1e

PR #87, opened 2019-09-20
Move to UFO sources and TTX dump for VTT data

### LINKED ISSUE #78: Suggestion: Use UFO format for source storage
# Description of the new feature/enhancement (with images if possible)

Currently, the font source is provided as a .glyphs file and a TTF binary with VTT code inside. This probably works fine for the designer, but both formats are not made for version control. There is another way: use the UFO format for source storage, a TTX dump for VTT code storage and put that in version control. Glyphs files and VTT binaries can then be generated.

Granted, the UFO format has its share of problems, but it is the only format explicitly made for version control. At my company, we use https://pypi.org/project/glyphsLib/ to round-trip from UFO to Glyphs and back. The UFOs are compiled with https://pypi.org/project/fontmake/, VTT hinting applied with https://github.com/daltonmaag/vttLib/.

Quick drive-by modifications to the UFOs could even be made with free tools like FontForge and TruFont.

# Proposed technical implementation details (optional)

You can have a look at https://github.com/daltonmaag/ubuntu and https://gitlab.gnome.org/GNOME/cantarell-fonts/ to see what this can look like.

The workflow is roughly

1. `ufo2glyphs CascadiaCode.ufo`
2. Work on the Glyphs file
3. `glyphs2ufo CascadiaCode.glyphs` (creates a Designspace file after the first time, which contains more information and is stored next to the UFO)
4. Look through changes to UFO and commit desired changes.

Compiling the font is done with `fontmake -u CascadiaCode.ufo -o ttf` (or `-m CascadiaCode.designspace`). Note that the cubic-to-quadratic curve conversion algorithm used has an effect on the VTT hinting and can completely destroy it. I wrote a hint-transplanter for this reason, have it somewhere. Hinting is injected with `python -m vttLib mergefile input.ttx output.ttf && python -m vttLib compile --ship output.ttf`, where input.ttx is obtained by `python -m vttLib dumpfile CascadiaCode-Regular-VTT.ttf input.ttx`.

The VTT workflow is messy, but only because VTT remains a sore thumb in font production 😢

madig:

I couldn't contain myself and went ahead on https://github.com/microsoft/cascadia-code/issues/78.

The build script should also include the glyph width modification script that the designer seems to run manually to cut a release, noted in the notes.

 
## References

GitHub Actions may be used to automatically do https://github.com/microsoft/cascadia-code/issues/57.

## PR Checklist
* [x] Closes #78

The motivation for this is detailed in #78.

I have additionally outfitted the UFO with Postscript blue zone information so I can produce OTFs and psautohint them:

grafik

VTT data is added as-is but is broken because the code converts outlines differently from Glyphs.

## Validation Steps Performed

- [ ] Update VTT code: the build will fail because the glyph order changed, the different outline conversion causes the VTT code to be outdated. Nice explosions though.
- [ ] Compare TTX dumps afterwards to look for changes outside hinting.


mdtauk:
Is there a way to add support for generating FontLab source files?

madig:
Fontlab has a vfb2ufo utility: https://blog.fontlab.com/font-utility/vfb2ufo/.

madig:
Rebased on master.

DHowett-MSFT:
The folder names here weird me out a little bit. What's up with them? They're not dates exactly... but they're also not not dates.

image

madig:
This is one of the impedance mismatches between how Glyphs does layers and how UFO handles them. In Glyphs, glyphs have layers. In UFO, layers have glyphs. This results in slightly different low-level semantics, but glyphsLib translates this losslessly, and only the glyphs in the `glyphs/` directory participate in font making. See these layers as extra work space for the designer.

madig:
@cinnamon-msft I just read your roadmap branch changes. Exciting :) One thing that you have to keep in mind currently for planning is that the chosen workflow affects hinting, or rather, what is hinted. Since the design outlines are cubic, TTFs allow only quadratic curves, and VTT works on those quadratic outlines directly, the chosen algorithm for converting the curves and the VTT code are intertwined. Currently, this branch produces different TTFs from what Glyphs outputs (uses some version of [Adobe's AFDKO](https://github.com/adobe-type-tools/afdko) under the hood), so the existing VTT code is completely broken. This is becomes relevant every time the conversion algorithm changes (be it by changing libraries/compilers or due to newer versions doing things differently -- this has happened before).

Your planning should include either engineering effort to use/match what the AFDKO does or allocate resources for adapting the hinting. I happen to be sitting on code that converts VTT code from old outlines to new ones and I'm allowed to open-source it, but it currently does not handle the new `Res*` commands and it does not handle TrueType assembly, so composites have to be redone manually.

madig:
Note to self: test if new UFO export from Glyphs 2.6.2 is good enough so we don't need glyphsLib here.

aaronbell:
As this is subsumed in the #181 update, closing :)

Issue #137, opened 2019-10-03
Cascadia Code Font must be reinstalled after reboot


dkgreer52:
I've been enjoying the new Cascadia font in Visual Studio 2017 and 2019, but I've discovered that each time I reboot Windows 10 (v1903), I have to reinstall the font to get it to work.  When I reinstall it, Windows says the font is already installed and asks me to confirm the reinstall.  After the reinstall, the font works fine again.

DanielKOcean:
I face the same problem, but the font sometimes disappears even without rebooting. Re-installation and reloading VSCode or VS solves the problem, but WTF?

DanielKOcean:
> I've been enjoying the new Cascadia font in Visual Studio 2017 and 2019, but I've discovered that each time I reboot Windows 10 (v1903), I have to reinstall the font to get it to work. When I reinstall it, Windows says the font is already installed and asks me to confirm the reinstall. After the reinstall, the font works fine again.

Hi, seems that I solved the problem.
1) (Not sure, that this action is needed) In "Properties" -> check "Unblock" checkbox, click "Apply" or "Ok" button, it will remove security mark from a font file;

![Unblock Font File](https://user-images.githubusercontent.com/9781016/68371761-729b1900-0150-11ea-82b4-ce0640713a4f.png)
2) In file drop down menu click "Install for all users" (see attached pic).
![Install Font](https://user-images.githubusercontent.com/9781016/68371441-c5c09c00-014f-11ea-82bb-03ae94dfe581.png)

aaronbell:
Sounds like this issue has been resolved. If you still have troubles, please feel free to re-open!

mick-d:
This is issue is still happening for me with "Cascadia Code PL". The PL version keeps being removed at reboot.

I have the Windows build 19041.329

grtrout:
I had this problem and tried various different things to resolve it. Eventually, I stumbled on a solution that worked. Rather than copying the files to the Fonts directory or using any of the Windows settings dialogues, simply right-clicking on the font files (e.g., *.ttf) and selecting "**Install _for all users_**" did the trick. I've now restarted my computer several times and not seen this issue.

mikhey:
I just submitted a ticket to the IT team at work b/c I thought it was McAfee attempting to "help" me by removing it.  In either case, this is infuriating to say the least!  - Windows 10 1909 Build 18363.1440

I don't have those options in the screenshot shown above.

dkgreer52:
[...]

PR #181, opened 2019-11-12
November Update

### LINKED ISSUE #33: Add glyphs for assorted additional characters
# Description of the new feature/enhancement (with images if possible)

Please add glyphs for accented characters (specifically, for my case, ISO-8859-1-equivalent; i.e., 
U+00C0 to U+00FF); the Greek alphabet, uppercase and lower (U+0370-03FF); and the copyright and trademark symbols (U+00A9 and U+2122).

I have more than a few filenames which contain the first; a considerable amount of mathematical code which makes use of Greek symbols in variable and function names; and, of course, copyright and trademark symbols appear repeatedly in comments and embedded messages.

### LINKED ISSUE #80: No visibility for = sign on terminal
Environment

Cascadia Code version number: 1909.16
Application (with version) used to display text: Linux Default Terminal
OS platform and version: Ubuntu
Screen resolution: 1366x768 pixels (96 dpi)

Any other software?
Steps to reproduce

cascadia font

Install cascadia font in the environment mentioned above you will see.
Expected behavior
Equal sign should be visible
Actual behavior
Equal sign in not visible at all

### LINKED ISSUE #117: Support for `≡`?
# Environment

```none
Cascadia Code version number: 1906.16
Application (with version) used to display text: Windows Terminal Preview
OS platform and version: Windows 10 Insider Preview 18990
Screen resolution (i.e. 220dpi): 96DPI

Any other software? N/A
```

# Steps to reproduce

1. Set font as Cascadia Code in Windows Terminal
2. Copy-paste `≡` multiple time in `pwsh`, `cmd` or even `ubuntu`. (The character is used in [posh-git](https://github.com/dahlbyk/posh-git))
3. Try moving cursor between `≡`s

# Expected behavior

It should be half-width character as in Consolas.

# Actual behavior

It's rendered as full-width character and confuses Windows Terminal. (Related with https://github.com/microsoft/terminal/issues/42 ?)

### LINKED ISSUE #118: fontconfig does not recognize Cascadia as 'monospace'
# Environment

```none
Cascadia Code version number: 1909.16
Application (with version) used to display text: GNU fontconfig
OS platform and version: Ubuntu 19.04 on GNOME 3.32
fontconfig 2.13.1
```

# Steps to reproduce

I am using `fonts.conf` as:

```xml
  <alias>
    <family>monospace</family>
    <prefer>
      <family>Cascadia Code</family>
    </prefer>
  </alias>

```

after updating the fontconfig and running `fc-cache -rfv`, Cascadia Code is listed as a new font however fc-query does not match the font with `monospace`:

```bash
$ fc-match monospace
DejaVuSansMono.ttf: "DejaVu Sans Mono" "Book"
```

where trying to query the font file:

```bash
$ fc-query ~/.fonts/Cascadia.ttf 
Pattern has 25 elts (size 32)
	family: "Cascadia Code"(s)
	familylang: "en"(s)
	style: "Regular"(s)
	stylelang: "en"(s)
	fullname: "Cascadia Code"(s)
	fullnamelang: "en"(s)
	slant: 0(i)(s)
	weight: 80(f)(s)
	width: 100(f)(s)
	spacing: 100(i)(s)
	foundry: "SAJA"(s)
	file: "~/.fonts/Cascadia.ttf"(s)
	index: 0(i)(s)
	outline: True(s)
	scalable: True(s)
	charset: 
	0000: 00000000 ffffffff ffffffff 7fffffff 00000000 00000000 00800000 00800000
	0020: 33000000 00000000 00000000 00000000 00000000 00000000 00000000 00000000
	0022: 00040000 00000000 00000000 00000000 00000000 00000000 00000000 00000000
(s)
	lang: fj|ho|ia|ie|io|nr|om|so|ss|st|sw|ts|uz|xh|zu|kj|kwm|ms|ng|rn|rw|sn|za(s)
	fontversion: 125118709(i)(s)
	capability: "otlayout:DFLT otlayout:latn"(s)
	fontformat: "TrueType"(s)
	decorative: False(s)
	postscriptname: "CascadiaCode"(s)
	color: False(s)
	symbol: False(s)
	variable: False(s)

```

The expected behavior as common with fontconfig is that when `monospace` is replaced with a specific font, any application using `monospace` would get `Cascadia Code` as the actual font.

The expected behavior is there for example with `Fira Code`. In addition, if I manually set the monospace font to Cascadia Code, it works. So, maybe an issue with fontconfig or how to set the name in `fonts.conf`?

### LINKED ISSUE #130: Feature Request: More Braille Characters?
# Description of the new feature/enhancement

I have in recent years been frustrated with the fact that I cannot use braille characters in the default windows terminals, but I am really excited to see that this font supports braille.  I immediately went to the [blessed-contrib repository](https://github.com/yaronn/blessed-contrib), a NodeJS module that uses braille characters for displaying dashboard information, and ran the demo file, it did mostly work.
blessedContribDashboard

It seems there are a few glitches which I suspect is because not all braille characters are supported.

If these defects could be fixed that would be tremendously helpful.

If you would like to run the demo for yourself, you can check out the [blessed-contrib repository](https://github.com/yaronn/blessed-contrib) and look at the instructions to run the demo, it's pretty straight forward.

### LINKED ISSUE #142: Missing Glyph U+25A0 (Black Square) and some other high-ASCII characters from CP437
# Environment

```none
Cascadia Code version number: 1910.040
Application (with version) used to display text: Windows Terminal & ConHost
```

The font is missing the U+25A0 (Black Square) glyph. In Windows Terminal it gets replaced by a full-width version from some other font, while in ConHost it shows as the missing empty rectangle.
This character is important as it is part of the base MDA/CGA / Codepage 437 characters set, and is often used along the full block and half blocks as a centered half block, and therefore should be included as part of the new box drawing glyphs.

Also, providing all 256 characters from CP437 would highly improve compatibility with original PC console/terminal outputs, so the other characters missing in the screenshot below would be a great addition.

Consolas:
MDA charset Consolas

Cascadia Code:
MDA charset Cascadia Code

### LINKED ISSUE #144: Feature Request: Inequality symbols with horizontal lower bar
# Description of the new feature/enhancement (with images if possible)

The symbols for "greater or equal" and "less or equal" have the lower bar angled, i.e. parallel to the greater than or less than symbol.  For some reason, most if not all of the ligature-enhanced fonts I've seen for programming do this, instead of having the lower bar parallel.

I do not claim to be an authority on this matter, but I've never seen this in mathematics or programming textbooks, or anywhere else other than programming fonts, really.  Indeed, these symbols ≥, ≤ render with horizontal lower bars with my browser defaults.  The angled bars must be a preference of enough people to be common, but it's very jarring when I see it.

image

### LINKED ISSUE #161: Missing glyphs cause display issues with "Show Invisibles" in BBEdit
# Environment

```none
Cascadia Code version number: 1910.04
Application (with version) used to display text: BBEdit 13.0.1
OS platform and version: macOS 10.14.6 (18G103)
Screen resolution (i.e. 220dpi): n/a

Any other software?
```

# Steps to reproduce

1. Download and install BBEdit 13.0.1. It is available from the BBEdit downloads page: <https://www.barebones.com/support/bbedit/updates.html>. (If you are running an older version of macOS, you can use an older version of BBEdit.)

2. Open a text file which contains spaces, nonbreaking spaces, and tabs.

3. Click the small gear above the upper left corner of the edit view to open the Text Options popover.

4. Turn on "Show Invisibles." Make sure that "Show Spaces" is turned on as well.

# Expected behavior

=> Invisibles placeholders shown for spaces, tabs, and line breaks. (Attached screen shot using Inconsolata for reference.)

Screen Shot 2019-10-21 at 9 51 00 AM

# Actual behavior

=> "missing glyph" placeholder shown for tabs

Screen Shot 2019-10-21 at 9 52 04 AM

BBEdit uses the following code points for invisibles glyph substitution:

0x00B7,		//	space -> dot
0x2022,		//	nbsp -> bullet
0x2206,		//	tab -> delta
0x00AC,		//	CR -> not sign
0x00B6,		//	FF -> paragraph
0x007C,		//	zero width -> vertical bar
0x00BF,		//	all others -> inverted question mark

### LINKED ISSUE #168: — and - look the same
# Environment

```none
Cascadia Code version number: 1910.04
Application (with version) used to display text: Any, e.g., Jupyter notebook 1.0.0
OS platform and version: Windows 10
Screen resolution (i.e. 220dpi): 1920x1080

Any other software? No
```

# Steps to reproduce

— and - look the same

image

# Expected behavior

They shouldn't look the same.

# Actual behavior

They look the same.

### LINKED ISSUE #176: Include the √ character (charcode 0x221A)
# Description of the new feature/enhancement (with images if possible)

Noticed when I ran the command `yo doctor` that their checkboxes were missing. This is the `√` character (charcode 0x221A). Consider including?

### LINKED ISSUE #177: Check mark symbol
# Description of the new feature/enhancement

The check mark symbol `✓` (U+2713) would be useful to have.

There are variations to consider (probably not all are necessary)
* ✓ check mark (U+2713)
* ✔ heavy check mark (U+2714)
* 🗸 light check mark (U+1F5F8)
* ⍻ not check mark (U+237B)
* ☑ ballot box with check (U+2611)
* ✅ white heavy check mark (U+2705)

### LINKED ISSUE #179: Feature Request: support "Geometric Shapes"
As a complement to the ["box drawing"](https://unicode-table.com/en/blocks/box-drawing) and ["block elements"](https://unicode-table.com/en/blocks/block-elements) characters, the ["geometric shapes"](https://unicode-table.com/en/blocks/geometric-shapes)(`U+25A0-25FF`) block would be a useful addition.

[`tig`](https://github.com/jonas/tig), specifically, uses characters from that block to display `git` log graphs.

### LINKED ISSUE #182: Unicode Character 'LINE SEPARATOR' (U+2028) support
# Steps to reproduce

'LINE SEPARATOR' (U+2028) is not supported and rendered as ? or L sep symbol

# Expected behavior

'LINE SEPARATOR' (U+2028) is rendered as Whitespace

# Actual behavior

'LINE SEPARATOR' (U+2028) is not supported and rendered as ? or L sep symbol

### LINKED ISSUE #183: Tab character is displayed with 'A' in BBEdit
# Environment

```none
Cascadia Code version number: 1910.04 (and with November Update #181 builds)
Application (with version) used to display text: BBEdit 13.0.3
OS platform and version: macOS 10.15.1
Screen resolution (i.e. 220dpi): n/a
```

# Steps to reproduce

1. Download and install BBEdit 13.0.2 from https://www.barebones.com/support/bbedit/updates.html
2. Create a new text document in BBEdit
3. Enter a `tab` character (0x0009)

# Expected behavior

The tab character should appear as ‘clear’. For reference, using Consolas:

consolas-tab-example

# Actual behavior

Using Cascadia Code (including the new November update #181 and the Mono variants) this is displayed thusly:

cascadia-tab-example

Note that the Mono variants display an accented A (Á) in the tab character.

### LINKED ISSUE #184: Use a 'pointing in' glyph for *true* single and double quotation marks
Open quotation marks, single and double, feel as though they’re pointing the wrong direction.

The characters have a bit more weight at the bottom but ‘point outward’, making it appear backwards.

Cascadia Code

Contrast this with Courier and Courier New, both of which use a similar glyph style:

Courier has a weighted bottom, pointing ‘in’:
Courier

Courier New has a weighted top, pointing ‘in’:
Courier New

***

Cascadia should use a ‘pointing inward’ glyph for its quotation marks.

I find the ‘weighted top’ style to be easier to read when using this style of glyph. For example, compare to SF Mono, whose quotes are very difficult to discern at anything but large sizes because it uses a ‘weighted bottom’ style:

SF Mono

aaronbell:

This PR covers extensive modifications to the Cascadia Code repository, including a switch of source format to UFO (with all necessary build files) and extensive character set extension to include a wide variety of characters, as well as Greek, Cyrillic, and Vietnamese. 

This pull request will replace the existing code base with a UFO-driven codebase, and includes a build.py file that generates the font and adds a VTT source. This PR encapsulates #87. 

## PR Checklist
* [x] Closes #184 - quote mark design modified
* [x] Closes #183 - glyph added
* [x] Closes #182 - glyph added
* [x] Closes #179 - glyphs added (thanks @mdtauk!)
* [x] Closes #177 - glyph added
* [x] Closes #176 - glyph added
* [x] Closes #168 - dash lengths modified for better differentiation
* [x] Closes #161 - glyphs added
* [x] Closes #144 - Greater & Equal / Less & Equal shapes modified
* [x] Closes #142 - codepage 437 and 850 now supported
* [x] Closes #130 - Braille characters added
* [x] Closes #117 - glyphs added
* [x] Closes #80 - version without ligatures created
* [x] Closes #33 - Greek / Cyrillic / Vietnamese support added. All requested languages covered

* [x] Requires [/images/cascadia-code-characters.png](/microsoft/cascadia-code/blob/master/images/cascadia-code-characters.png) to be updated

This update includes a major update to the font's build process and character coverage. In terms of build process, the font now is built from a central UFO source, and outputs VTT source files. Additionally, Powerline only, and full NerdFonts variants are created as well (Cascadia Code PL and Cascadia Code NF, respectively), as well as versions of each with the ligatures disabled. 

In terms of character coverage, the font now includes the core Greek and Cyrillic code blocks (including localized Cyrillic forms), Vietnamese character support, an extensive set of numerator / denominator figures, currency symbols, a variety of common math symbols, geometric shapes, and other symbols. Other modifications include updates to the dash lengths, modified ligature for greater / equal & lesser / equal, and a variety of other small fixes. 

## Validation Steps Performed
Font reviewed in Font Bakery to ensure the font file functionality is correct. There are several warnings / fails related to the font metrics / metadata, but nothing that affects font functionality.


madig:
I spy an added " .DS_Store" file 😱

Also, please add the blue zone data from https://github.com/microsoft/cascadia-code/blob/20d03e2f08cb747eb2210c1aab11f094a0b01cf5/sources/CascadiaCode-Regular.ufo/fontinfo.plist#L235-L268 so I can make deliciously autohinted OTFs ☺

aaronbell:
Still a draft! :P

madig:
> `instanceNF = generator.generate_instance(instance_descriptor)  #rebuilding these instead of as a deepcopy avoids an error when compiling the TTF`

What error does it throw?

build.py
```diff
@@ -0,0 +1,124 @@
...
+
+        #4.2 GET BACK TO WORK
+
+        NF_UFO = ufoLib2.objects.font.Font.open(INPUT_DIR / "nerdfonts" / "NerdfontsNF.ufo")
+        PL_UFO = ufoLib2.objects.font.Font.open(INPUT_DIR / "nerdfonts" / "NerdfontsPL.ufo")
+        
+        #4.5 Adding glyphs
+        print ("Adding Nerd Font glyphs")
+        for glyph in NF_UFO.glyphOrder:
+            if glyph in instanceNF.glyphOrder:
```
madig: Try

```python
            if glyph not in instanceNF.glyphOrder:
                instanceNF.addGlyph(NF_UFO.get(glyph))
```

build.py
```diff
@@ -0,0 +1,124 @@
...
+        # 5. Generate non-Ligature versions
+
+        instance_noLIG = copy.deepcopy(instance)
+        instanceNF_noLIG = copy.deepcopy(instanceNF)
+        instancePL_noLIG = copy.deepcopy(instancePL)
+
+        with open(INPUT_DIR / "features" / "features.fea", 'r') as feaCode:
+            noLIG_fea = feaCode.read()
+
+        setattr(instance_noLIG.features,"text",noLIG_fea)
```
madig:
```suggestion
        instance_noLIG.features.text = noLIG_fea
```

build.py
```diff
@@ -0,0 +1,124 @@
...
+        setattr(instanceNF_noLIG.features,"text",noLIG_fea)
+        setattr(instancePL_noLIG.features,"text",noLIG_fea)
+
+        setattr(instance_noLIG.info, "familyName","Cascadia Code N")
```
madig:
```suggestion
        instance_noLIG.info.familyName = "Cascadia Code N"
```

madig:
Also, this just occured to me: the script that I made is ready for interpolation, which is why I used the `fontmake.instantiator` module. If interpolation is not planned anytime soon, the build script can probably be simplified (i.e. you can set PANOSE directly in the UFOs).

aaronbell:
@madig, thanks for the feedback and suggestions for the build script! I definitely thought about simplifying the build script, but I think the plan is to add weight variants early next year, so might as well leave it as is if it works :)

aaronbell:
> > `instanceNF = generator.generate_instance(instance_descriptor)  #rebuilding these instead of as a deepcopy avoids an error when compiling the TTF`
> 
> What error does it throw?

Assertion Error. I couldn't sort out exactly what was happening. Essentially when I did a deep copy, then tried to modify the UFOs and compile the TTFs, I got these endless assertion errors. It seemed to me that it was related to trying to compile the same data twice, or something. Not sure. Anyway the current implementation works, which is what I care about :)

DHowett-MSFT:
Line spacing looks smaller on the build output of this versus the original Glyphs version!

## orig

image


## new

image

aaronbell:
@DHowett-MSFT Ah! This is why it is still a draft. :). Not fully done!

This must be because I turned on the "useTypoMetrics" flag, which instructs the rasterizer to render the text using the actual metrics of the font, rather than the (somewhat) arbitrary winAscent and winDescent values. However, the downside is that then it is on the application to increase the line spacing appropriately.

Looking at the sample images, I think I'll disable it as most applications where this font is being used likely won't have support for modifying line-height. And the above is not a great look. :)

madig:
This might be an opportunity to work on Terminal's line height stuff. Note how the spec recommends turning use typo metrics on for new fonts: https://docs.microsoft.com/en-us/typography/opentype/spec/os2#fsselection. I suppose terminal is not rendering the line gap?

aaronbell:
> This might be an opportunity to work on Terminal's line height stuff. Note how the spec recommends turning use typo metrics on for new fonts: https://docs.microsoft.com/en-us/typography/opentype/spec/os2#fsselection. I suppose Terminal is not rendering the line gap?

Totally! Offering the ability to customize letter spacing and line height would be great for accessibility purposes. But I'm more concerned about other applications that don't provide this kind of flexibility and many folks coming here to file bugs about tight line spacing. :)

madig:
Let's fix all the applications!

DHowett-MSFT:
@cinnamon-msft should the non-ligatures version of the font be named `Cascadia Mono`? It's got a nice ring to it, and it makes it feel like a whole family of fonts.

mdtauk:
Cascadia Mono
Cascadia Code
Cascadia Power
Cascadia Nerd

cinnamon-msft:
`Cascadia Mono` is a winner :)

aaronbell:
Yeah, I'm planning:
`
Cascadia Code
Cascadia Code NF
Cascadia Code PL
Cascadia Mono
Cascadia Mono NF
Cascadia Mono PL
`

reflectronic:
The glyphs for `>=` and `<=` are way too wide now. I think it would be better to keep them as they were previously and add a stylistic set to keep the line straight, like Fira Code does.

aaronbell:
@reflectronic Fair :). I felt the original was too narrow, so I'll find a middle ground.

mdtauk:
I thought all glyphs were in multiples of the single glyph width?  Wouldn't a ligature be the width of however many glyphs form it?

aaronbell:
@mdtauk Yes-ish. The `>=` and `<=` were actually narrower—about a single width within a double-width space. So I modified them to be more full double width. However, I agree that it starts to feel really wide (versus with the `->` ligatures where the horizontal line can just extend further and the `>` is kept fairly constrained). So finding a middle ground will let the `<` look more normal while using the space more fully. 

I also think switching to a horizontal bar helps with the legibility (versus the matching bar). So I'd rather keep that :)

DHowett-MSFT:
Should we set `openTypeOS2Type` to `0`[*]? I wonder if it's the cause of things like #137, since technically the font isn't licensed for local installation.

_[*]: 0: Installable embedding: the font may be embedded, and may be permanently installed for use on a remote systems, or for use by other users. The user of the remote system acquires the identical rights, obligations and licenses for that font as the original purchaser of the font, and is subject to the same end-user license agreement, copyright, design patent, and/or trademark as was the original purchaser. ([source](https://docs.microsoft.com/en-us/typography/opentype/spec/os2#fstype))_

DHowett-MSFT:
It looks like you did the same shine reduction you did on `C` and `S` on a couple other glyphs, but the 3 looks a bit more different than the others. Intended?

image

(let me know if I should stop asking questions about WIP things :smile: I'm just excited!)

aaronbell:
Haha. No, feel free to ask away!

Yes, the changes are intentional. I actually rather liked the way the /3 and /2 (and others) resolved, but the more I kept working on related shapes, the more I felt that the flat terminal no longer had a place in the design. 

In the case of the /3, the bottom also had to change in order to accommodate the terminals like on the /S and /C. So it feels like a more significant change. And then with all that, the sidebearings had to be re-adjusted.

build.py
```diff
@@ -0,0 +1,118 @@
...
+        instanceNF = generator.generate_instance(instance_descriptor)  #rebuilding these instead of as a deepcopy avoids an error when compiling the TTF
+        instancePL = generator.generate_instance(instance_descriptor)
+
+
+        # 4.1 Modifying some attributes to correct the metadata
+        setattr(instance.info,"openTypeOS2Panose",[2,11,6,9,2,0,0,2,0,4])
+        setattr(instanceNF.info,"openTypeOS2Panose",[2,11,6,9,2,0,0,2,0,4])
+        setattr(instancePL.info,"openTypeOS2Panose",[2,11,6,9,2,0,0,2,0,4])
+
+        instanceNF.info.familyName = "Cascadia Code NL"
```
DHowett: This seems like it should be `Cascadia Code NF`

madig:
> We may want to not ship the binary ttf files inside the repo any more.

Yes, releases should be done via tagging a commit and having CI append the "assets" to the GitHub release page.

aaronbell:
> This deletes the entire .github folder with all the issue templates

@DHowett-MSFT This deletes the issue templates? I'm not seeing any of them enumerated anywhere in the repro to preserve. I guess one approach is to copy them elsewhere, make the update, and then re-integrate?

DHowett-MSFT:
@aaronbell so on OS X you'll need to show hidden folders :smile: it's in a folder called `.github`

DHowett-MSFT:
About the conflicts: I'm not concerned about the text files, but i AM about the glyphs/ttf. We're generally 100% certain that the font in this PR matches the font that's currently in the repo, right?

(A delete will only conflict if git thinks you're deleting a file that _doesn't match_ the one in the repo, indicating perhaps a missing commit or broken state)

aaronbell:
> We're generally 100% certain that the font in this PR matches the font that's currently in the repo, right?

The .glyphs file, and .ttf file that are in the repro are source files. The glyphs file is now being represented by the .ufo and .designspace files in the 'build' folder. Similarly, the .ttf is a VTT source that we have removed from the repro and is now represented by the ttx hinting file, and is what is exported by the build.py flow.

I will say that they are not 1:1 exactly the same because the new version contains the Greek / Cyrillic bits and bobs. But yes, I feel confident that the existing source files are being properly represented in the new version.

madig:
> We're generally 100% certain that the font in this PR matches the font that's currently in the repo, right?

Run fontTools' ttx on both and diff the output. The binary will be different if produced by the script instead of Glyphs, but not every diff is meaningful. 

Also, blue zones please 😊

DHowett-MSFT:
@madig I'd love to review those as a separate PR on top of this otherwise really quite huge one, if that's alright :smile:

madig:
Uh. The build script doesn't work?

```
build.py:47:17: F821 undefined name 'instanceNF'
```

I see we need some CI around here :)

aaronbell:
Weird. I ran the script and it worked previously. Wonder why. Anyway, will comment that out.

madig:
Ok :D I may work on that script later.

Aside: I just now remembered that I recently stumbled over https://github.com/typemytype/booleanOperations/issues/55. That's the default overlap-removal library used in the build pipeline. And the contour order of various contour-intensive glyphs like `shadelight` partly depend on the platform you are compiling on. :scream: that does not happen with the other library pathops (which uses Skia code), but that may change contour order and other things, invalidating hinting. This is a real PIA issue and basically means that the CI should run on the platform you originally compiled the fonts on. MacOS I presume?

DHowett-MSFT:
FWIW I’m also working on a script cleanup :)

madig:
Please run https://github.com/psf/black/ on the script :grin:

madig:
Also: the ligature widths are not being corrected. You need this code: https://github.com/microsoft/cascadia-code/pull/87/files?file-filters%5B%5D=.py#diff-2955d5257f635de1df53f55a171ca5c7R15-R36

aaronbell:
What do you mean the ligature widths are not being corrected? They are not intended to be expanded to 2x, 3x, 4x widths, but remain at 1x width (1200funits) and use a spacing character.

madig:
Oh? Ok sorry, I missed that. I was going from the code in the font note saying how to produce production glyphs.

aaronbell:
No worries. Because working on glyphs that have negative sidebearings is kind of a PITA, I wrote a script to convert between production mode (1200 width) and edit mode (whatever multiple was intended). I think you may have confused which is which :)

madig:
@DHowett are you also adding a CI script? I made a GitHub Actions one at https://github.com/madig/cascadia-code/blob/20d03e2f08cb747eb2210c1aab11f094a0b01cf5/.github/workflows/pr.yml

ExE-Boss:
Can&nbsp;we&nbsp;get the&nbsp;`Cascadia*.ttf` files&nbsp;added to&nbsp;the&nbsp;[`v1911.20`](https://github.com/microsoft/cascadia-code/releases/tag/v1911.20)&nbsp;tag?

aaronbell:
The release is still in draft mode. It'll likely be made available today.

onovotny:
I don't see the Nerd Font variants published that are indicated in the notes above?

DHowett-MSFT:
Indeed. We're still ironing out some questions we have about some of the glyphs.

PhMajerus:
This version does not yet support codepages 437 and 850 completely.
I guess I should have included more details in #142, but the missing characters are not easy to generate with easy to reproduce steps in cmd.exe.

The last missing glyphs are in the characters for values 0x01 to 0x1F, which are typically used as C0 control characters, but can be output as characters by CUI apps that directly control the screen buffer.

CP437-850 missing glyphs

So for complete compatibility with older CUI apps, the following characters are also required:
`
    0123456789ABCDEF
0_   ☺☻♥♦♣♠•◘○◙♂♀♪♫☼
1_  ►◄↕‼¶§▬↨↑↓→←∟↔▲▼
`
Their Unicode codepoints are as follows:
0_ : (NULL), U+263A, U+263B, U+2665, U+2666, U+2663, U+2660, U+2022, U+25D8, U+25CB, U+25D9, U+2642, U+2640, U+266A, U+266B, U+263C
1_ : U+25BA, U+25C4, U+2195, U+203C, U+00B6, U+00A7, U+25AC, U+21A8, U+2191, U+2193, U+2192, U+2190, U+221F, U+2194, U+25B2, U+25BC

As far as I can tell, the characters in 0x01, 0x02,, 0x0E and 0x0F are missing, codepoints U+263A, U+263B, U+266B and U+263C, which are the two smiley faces, the double music note and the sun.

It would be great to have these included for full console compatibility.

aaronbell:
@PhMajerus Shoot! It looks like the list I pulled for CP437 was not fully complete (it didn't include all the relevant symbols). I'll reopen #142 with the list of unicodes missing. 

For the control characters—I've never actually included those in a font—and am not sure I fully understand their purpose. Can you provide more info? Thanks!

PhMajerus:
@aaronbell You can leave #142 closed as I didn't include much details about the 0x01 to 0x1F range besides "providing all 256 characters from CP437 would highly improve compatibility with original PC console/terminal outputs" in that bug report.
I now moved the details for the missing characters to #205 to avoid the long post here.

About the C0 control characters, these are not glyphs in a font, and not characters in the ASCII standard, but in-band control for CUI apps.
When a CUI app outputs text, the 0x01 to 0x1F characters are typically not displayed, but are instead interpreted by the console as being control commands such as bell (make the terminal make a sound), backspace, tab, return, ... as well as a bunch of commands or data separators meant for serial communication.
This means they do not display anything, and instead make the console or the serial device change some state.
However, MS-DOS apps could also write values directly into the text screen buffer memory, so it was technically possible to use these values for characters that more advanced apps could display.

All this means that these characters would not be usable by CUI apps that simply write to the console like a stream, they are even less common than high-ASCII, but could be used by apps that want to control the whole screen buffer for text-based GUI apps for example. They are not part of ASCII, but were usable since the early IBM-PC days.

From a font point of view, you don't need to care about the C0 control characters, but only to know that these values have been reused by the original MDA and CGA adapters to provide a few more characters to apps that handle the text screen buffer directly, and therefore these glyphs need to be included in your set if you want your font to be compatible with the original console.

PR #188, opened 2019-11-20
Add Postscript blue zones for letters and numbers

madig:

Add Postscript blue zones that enclose letters and common numbers and define vertical and horizontal stem widths. This helps with autohinting OTFs.

## PR Checklist
* [ ] ~~Closes #xxx~~
* [ ] I've discussed this with core contributors already. If not checked, I'm ready to accept this work might be rejected in favor of a different grand plan. ~~Issue number where discussion took place: #xxx~~

When not producing TTFs but OTFs, hinting is a matter of running https://github.com/adobe-type-tools/psautohint/ on the resulting binary. For the autohinter to do its job, it needs to know the standard widths of a font plus so-called blue zones, which tell the tool which heights points of a glyph should be snapped to. They also help with consistency in the design phase, because e.g. Glyphs will indicate when a point is within or outside a blue zone.

TTFs are unaffected by this change since the format does not use them.

## ~~Before (if applicable) and~~ After Images of the Character(s)

Bildschirmfoto von 2019-11-20 20-49-41
Bildschirmfoto von 2019-11-20 20-49-45
Bildschirmfoto von 2019-11-20 20-49-50
Bildschirmfoto von 2019-11-20 20-54-16

## Validation Steps Performed

Manually looking at the resulting font in a terminal and in VSCode.

### DIFF
--- a/sources/CascadiaCode-Regular.ufo/fontinfo.plist
+++ b/sources/CascadiaCode-Regular.ufo/fontinfo.plist
@@ -205,7 +205,16 @@ THE FONT SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR 
     <key>openTypeOS2VendorID</key>
     <string>SAJA</string>
     <key>postscriptBlueValues</key>
-    <array/>
+    <array>
+      <integer>-20</integer>
+      <integer>0</integer>
+      <integer>1060</integer>
+      <integer>1080</integer>
+      <integer>1420</integer>
+      <integer>1440</integer>
+      <integer>1500</integer>
+      <integer>1520</integer>
+    </array>
     <key>postscriptFamilyBlues</key>
     <array/>
     <key>postscriptFamilyOtherBlues</key>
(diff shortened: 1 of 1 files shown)


mdtauk:
PS Hinting for OTF is separate to TTF hinting.

madig:
Yes. This is for my private OTFing/in anticipation for making official OTFs one day. I love the way FreeType renders them.

DHowett-MSFT:
Hey @aaronbell -- since we're not producing our own OTFs, should we merge this to make it so that other people can? I feel bad having let this languish for so long.

aaronbell:
@DHowett-MSFT Sure. I haven't had a chance to review the additions, but I figure @madig got it right :). 

It might be worth considering building our own OTF files, though. ufo2ft also has a makeOTF function, so it shouldn't be too hard to add it into the build script.

madig:
I've been using these in my home-made OTF and everything looks great. So, unless someone changed fundamental design heights, they should be good.

DHowett-MSFT:
@madig thank you; I'm sorry it took so long to merge this. :smile:

madig:
No worries, this PR paid dividends in the meantime.

PR #190, opened 2019-11-21
Add a CI workflow for building pull requests

DHowett-MSFT:
Taken with attribution from @madig in #87 (he's the author of this commit :smile:)

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -5,6 +5,13 @@ on: [push]
 jobs:
   build:
 
+    # Build the fonts on macOS because we use the default overlap removal library
+    # booleanOperations, which unfortunately inherits
+    # https://sourceforge.net/p/polyclipping/bugs/195/ that manifests for certain
+    # glyphs like `shade` and `dkshade`. The contour order makes no visual
+    # difference, but is important for TrueType hinting, where instructions work on
+    # point indices.  The font was originally build on macOS and then hinted in VTT,
+    # so we have to use macOS to not potentially invalidate the hinting.
     runs-on: macos-latest
     
     steps:
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -1,0 +1,21 @@
+name: Build
+
+on: [push]
+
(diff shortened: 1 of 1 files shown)


madig:
Thanks! I'd add something like this:

```
    # Build the fonts on macOS because we use the default overlap removal library
    # booleanOperations, which unfortunately inherits
    # https://sourceforge.net/p/polyclipping/bugs/195/ that manifests for certain
    # glyphs like `shade` and `dkshade`. The contour order makes no visual
    # difference, but is important for TrueType hinting, where instructions work on
    # point indices.  The font was originally build on macOS and then hinted in VTT,
    # so we have to use macOS to not potentially invalidate the hinting.
    runs-on: macos-latest
```

Also @aaronbell: Where did you build the font that you hinted in VTT?

aaronbell:
@madig Huh. That's a fun bug. I did build the fonts on a Mac. 
Those two glyphs you called out, though, are actually not hinted, so that's a plus :)

madig:
Yes, I was baffled that no one found this sooner. It only affects _some_ complicated glyphs that have a lot of contours on the same x coordinate or something, so this is more a precaution than anything else.

DHowett-MSFT:
I'll add that comment. Thanks!

.github/workflows/pr.yml
```diff
@@ -0,0 +1,21 @@
+name: Build
+
+on: [push]
+
+jobs:
+  build:
+
+    runs-on: macos-latest
```
DHowett-MSFT:
```suggestion
    # Build the fonts on macOS because we use the default overlap removal library
    # booleanOperations, which unfortunately inherits
    # https://sourceforge.net/p/polyclipping/bugs/195/ that manifests for certain
    # glyphs like `shade` and `dkshade`. The contour order makes no visual
    # difference, but is important for TrueType hinting, where instructions work on
    # point indices.  The font was originally build on macOS and then hinted in VTT,
    # so we have to use macOS to not potentially invalidate the hinting.
    runs-on: macos-latest
```

PR #196, opened 2019-11-22
Adding GASP table data & removing from VTT ttx

### LINKED ISSUE #193: Cascadia Code PL renders poorly in VSCode
# Environment

```none
Cascadia Code PL 1911.20
Application (with version) used to display text: VSCode 1.40.1
OS platform and version: Windows 10
Screen resolution (i.e. 220dpi): various tried (see below)
```

# Steps to reproduce

On a monitor with 100% scaling (aka no scaling, I think?), Cascadia Code PL renders poorly. A monitor with scaling enabled renders the font nicely. 

image

# Expected behavior

It should render like the non-PL variant of Cascadia Code.

image

aaronbell:

This PR adds the GASP table to the build.py script to enable consistent rendering across the different iterations of the font. 

## PR Checklist
* [x] Closes #193 
* [x] CLA signed.

Code adds a gasp table that turns on smoothing, and enables hinting between 9ppem and 50ppem. 

## Validation Steps Performed

Built font with script and verified the correct GASP setting have been set.

### DIFF
--- a/build.py
+++ b/build.py
@@ -51,7 +51,7 @@ def build_font_instance(generator, instance_descriptor, *steps):
 
     setattr(instance.info, "openTypeOS2Panose", [2, 11, 6, 9, 2, 0, 0, 2, 0, 4])
 
-    instance.info._openTypeGaspRangeRecords =[
+    instance.info.openTypeGaspRangeRecords =[
         {
             "rangeMaxPPEM" : 9, 
             "rangeGaspBehavior" : [1,3]
--- a/build.py
+++ b/build.py
@@ -51,6 +51,21 @@ def build_font_instance(generator, instance_descriptor, *steps):
 
     setattr(instance.info, "openTypeOS2Panose", [2, 11, 6, 9, 2, 0, 0, 2, 0, 4])
 
+    instance.info._openTypeGaspRangeRecords =[
+        {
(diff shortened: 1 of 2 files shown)


build.py
```diff
@@ -51,6 +51,21 @@ def build_font_instance(generator, instance_descriptor, *steps):
 
     setattr(instance.info, "openTypeOS2Panose", [2, 11, 6, 9, 2, 0, 0, 2, 0, 4])
 
+    instance.info._openTypeGaspRangeRecords =[
```
DHowett-MSFT: I'm concerned that this property starts with an underscore -- it indicates it might be private. @Madig, are we safe to change this do y'know?

aaronbell: I can remove the underscore if it helps :). Apparently it works the same either way.

madig: The current version without underscores is correct. We use class properties (getters & setters) for some info attributes to automatically convert lists and dicts into the underlying data structures, the underscore attribute then holds the actual data.

aaronbell: (except that ufolib2 (and the UFO spec) had the GASP settings [enumerated wrong](https://github.com/fonttools/ufoLib2/pull/44) and so I bypassed the system entirely) 😄

madig: Good catch. My coworker suggested that we always use the flags together so we never noticed.

PR #259, opened 2020-05-01
Increase maxComponentCount (because VTTShell doesn't)

DHowett-MSFT:

I upped it to 4 and verified that VTT emits a hinted font with it set to 3. This will give us some breathing room until it next strikes. :smile:


DHowett-MSFT:
Sorry for the stray highlight, other-Aaron. This repository wants to load _the entirety of all Microsoft github accounts_ when I click "reviewers", and then the magic of GitHub's UI makes stuff jump around before I can click on it.

madig:
(Keep in mind that the TTX dump is read in via fontTools, which expects table dumps to be complete, so you can't just remove an attribute. You can however set values to zero.)

PR #260, opened 2020-05-02
Updating scripts for OTF and WOFF2

### LINKED ISSUE #236: Provide WOFF for Cascadia Code
Would it be possible to get a webfont (WOFF) for Cascadia Code?

Many technical bloggers use monospaced fonts when including code samples or examples of commands in articles. A webfont would make it easier to include in a blog or webpage.

Also, with the rise of online code editors like Eclipse Che, or online writing tools like StackEdit, being able to use Cascadia Code would be awesome!

aaronbell:
Updated the build.py script to also build OTF versions of the Cascadia Code fonts (in light of the blue zones added!). Additionally added a "build_WOFF.py" script that will process the TTF files in the "build" folder to produce WOFF2 files. Throws an error if VTT production tables present in the fonts. 

## PR Checklist
* [x] Closes #236 
* [x] CLA signed.

- Updated build.py to also build OTF file
- Added build_WOFF.py to process TTFs to create WOFF2 files (if VTT production files have been removed)
- Updated the Requirements.txt file to include compression library for WOFF2 generation


build.py
```diff
@@ -44,42 +44,51 @@ def _set(instance):
...
-    instance_font.save(file_path)
-
-    print(f"[{familyName}] Done: {file_path}")
+    for format in ["ttf","otf"]:
+        instance = generator.generate_instance(instance_descriptor)
+
+        for step in steps:
+            step(instance)
+
+        setattr(instance.info, "openTypeOS2Panose", [2, 11, 6, 9, 2, 0, 0, 2, 0, 4])
```
madig, line 53:
```suggestion
        instance.info.openTypeOS2Panose = [2, 11, 6, 9, 2, 0, 0, 2, 0, 4]
```

aaronbell:
@DHowett-MSFT I figured you can add the WOFF2 converter to your build script to make things easier for ya :D

build.py
```diff
@@ -44,42 +44,51 @@ def _set(instance):
...
+                    "rangeMaxPPEM" : 50,
+                    "rangeGaspBehavior" : [0,1,2,3]
+                },
+                {
+                    "rangeMaxPPEM" : 65535,
+                    "rangeGaspBehavior" : [1,3]
+                },
+            ]
+
+        familyName = instance.info.familyName
```
madig, line 71: Block can be shortened to
```
        file_stem = instance.info.familyName.replace(" ", "")
        file_path = (OUTPUT_DIR / file_stem).with_suffix(f".{format}")
```

build_WOFF.py
```diff
@@ -0,0 +1,20 @@
+import glob, os
+from fontTools import ttLib
+
+
+os.chdir("build")
+
+for file in glob.glob("*.ttf"):
```
madig, line 7: You could use pathlib.Path here, simplifies the code somewhat:
```python
from pathlib import Path

for file in Path(".").glob("*.ttf"):
    ttfile = ttLib.TTFont(file)

    if "TSI0" in ttfile:
        print(f"Warning! VTT production files present in font: {file}")
        print("Please ship the font out of VTT before converting")
        print("*** *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***")
        sys.exit(1)

    print(f"{file}: Generating WOFF2")
    ttfile.flavor = "woff2"
    ttfile.save(file.with_suffix(".woff2))
    print("All done")
    print("*** *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***") 
```

madig:
You may also want to run psautohint over the OTFs like in https://gitlab.gnome.org/GNOME/cantarell-fonts/-/blob/master/scripts/make-static-fonts.py#L38-39

build_WOFF.py
aaronbell: Looks good!

aaronbell:
Added psautohint functionality to the build script and also tried again to clean out the hints from those imported glyphs. I thought I'd gotten them before, but somehow snuck back.

aaronbell:
BTW, psautohint was kicking out a whole slew of warnings about 'bogus BBox" I assume partly because of the oddity of the ligatures. Anyway, I filed those away to a "log.txt" file in the build folder.

build.py
```diff
@@ -176,5 +178,10 @@ def build_font_instance(generator, instance_descriptor, *steps):
                     step_merge_nf,
                 )
 
+        print("Autohinting OTFs")
+
+        for file in Path("build").glob("*.otf"):
```
madig, line 183: This will rehint everything on every instance.

madig:
I think psautohint is a bit anal about that, but I don't know the real value of those warnings.

build.py
aaronbell: Yeah. Probably not the most efficient way to do it. But this way we don't need to be reliant on the uploader running psautohint before sending it up?

aaronbell:
It seemed to me that the hinting all went through despite the warnings—so I figured they were just for reference rather than any specific blocker.

build_WOFF.py
DHowett-MSFT: (I'm curious: this comment is marked resolved, but the script still isn't using pathlib. Shall we file a followup workitem to clean it up? :smile:)

aaronbell:
Derp. I totally forgot that it doesn't just automerge the suggested changes. Alright, I've gone through and fixed those up :)

build_WOFF.py
aaronbell: Merged it in (and modified a bit).

build.py
madig: I think you just need to dedent that block so that it is run after all instances have been written to disk :)

build_WOFF.py
```diff
@@ -0,0 +1,19 @@
+import sys
+from pathlib import Path
+
+from fontTools import ttLib
+
+for file in Path("build").glob("*.ttf"):
+    ttfile = ttLib.TTFont(file)
+
+    if "TSI0" in ttfile:
```
madig, line 9: Shouldn't this exit? It is an error after all and there will be no WOFF2 file, even though the next print statement says "all done".

aaronbell: I get super annoyed when scripts that are supposed to run on multiple files exits out after the first one. What if there’s more files that the error would occur on? I’d prefer to know it all at once rather than only see one at a time.

But I do agree that the strings could be better. :)

build.py
aaronbell: Ah! I see. It hasn't mattered up to this point because there's only the only instance in the font, but yes, it makes sense to dedent it so it doesn't repeat repeatedly.

PR #304, opened 2020-06-29
Make Cascadia Code a variable font w/ 6 named variations

### LINKED ISSUE #25: Any plans for a lighter weight? (Or am I using it wrong?)
Tried switching to CC in VS Code on macOS, from my previous font of SFMono-Regular, by installing the .ttf into Font Book, and putting `'Cascadia Code'` at the beginning of the Font Family box in VS Code. 

To my eyes, the font is a bit heavier than what I'd prefer. It seems as though only one weight was installed, "Regular," is that right? If so, any plans to provide lighter weights to choose from?

### LINKED ISSUE #43: Bug Report: parenthesis width changes when bold
# Environment

```none
Cascadia Code version number: 1909.16
Application (with version) used to display text: pyCharm
OS platform and version: windows 10, 18362.356
Screen resolution (i.e. 220dpi): multiple screens, 4k on 27" and FullHD 15"

Any other software? Issue not present in Notepad++ (but that editor does not display combined glyphs (e.g. !=) correctly so it may be irrelevant.
```

See attached images, anything right of ( gets pushed a few pixels when ( is highlighted. Same is true for [ and {

image
image

### LINKED ISSUE #46: == ligature
The distinction between "=" and "==" is quite subtle which could be dangerous for c++ code, maybe a 3 line equals for the "==" ligature could be more appropriate?

### LINKED ISSUE #90: Feature Request: Consider making it a Variable Font
# Make Cascadia Code a Variable Font

Additional weights, widths, should be generated by making the font family Variable.  This means all glyphs should have the same number of nodes defining the outline, between the light and bold weight axis, as well as between the narrow and wide width axis.

By making the typeface variable, it allows for users to adjust the base weight to aid in readability, whilst allowing for bold text for contextual styling and emphasis, whatever the chosen weight value.

### LINKED ISSUE #128: Feature Request: Provide more weights
# Description of the new feature/enhancement
Many IDE's do not have the settings as a configurable file( like some text editors: example Atom has stylesheet.less, Visual Studio Code has settings.json ) where you can choose the weight of the font.
I am thinking that more weights would be awesome to have. I really like the Iosevka structure:
It has  weights 8 weights but it would be more than enough to have 4 Light, Regular, Medium and Bold:
Thin
Thin italic
Thin oblique

Extralight
Extralight italic
Extralight oblique

Light
Light italic
Light oblique

Regular
Italic
Oblique

Medium
Medium italic
Medium oblique

Semibold
Semibold italic
Semibold oblique

Bold
Bold italic
Bold oblique

Extrabold
Extrabold italic
Extrabold oblique

### LINKED ISSUE #284: Use better `==` symbol
It is hard to distinguish between these two:
```
1==2
1=2
```

image

Borg-sans-mono has a better symbol for this:
https://github.com/marnen/borg-sans-mono
image

### LINKED ISSUE #285: Cascadia Mono shouldn't use contextual alternate for x
I was working with a third-party service that provided me with a URL of this form:

    https://example.com/1234x5678.png   (1 2 3 4 x 5 6 7 8)

I saw this URL in Windows Terminal, which defaults to Cascadia Mono. The way this URL displayed was this:

    https://example.com/1234×5678.png

This made me think the third-party service is using × (U+00D7 MULTIPLICATION SIGN) instead of the usual letter x (U+0078 LATIN SMALL LETTER X). This confusion is not something I want from a terminal. In my opinion, the contextual alternate should only be in Cascadia Code.

Here’s a sample (in Word, with ligatures and contextual alternates enabled):

image

And in Windows Terminal:

image

aaronbell:
##  Summary of the Pull Request

This major update adds a weight range to Cascadia Code. This font is now being built as a Variable Font, which enables users to select the perfect weight for their preference. 

## PR Checklist
* [x] Closes #25 - weight axis added
* [x] Closes #43 - with weight addition, parenthesis width is preserved 
* [x] Closes #284 - ligature now broken for easier recognition
* [x] Closes #90 - produced as a variable font
* [x] Closes #128 - MORE WEIGHTS 
* [x] Closes #285 - contextual code removed

* [x] CLA signed.

##Detailed Summary
The weight range now extends from ExtraLight (200) to Bold (700), with the current Regular set at 500. As a variable font, OS / rendering engine support may vary. Users running Windows 10 and Windows Terminal will have access to the full range of font weights. Other applications may only have access to the named instances (ExtraLight / Semilight / Light / Regular / SemiBold / Bold) depending on inbuilt support. 

Static instance OTFs are also provided. At current, static TTFs are not built in this update, but it is something we will consider in the future. 

Screen Shot 2020-06-29 at 12 32 31 PM

## Validation Steps Performed
Variation implementation tested on Windows and Mac, font hinted and reviewed on high and low DPI devices.


DHowett:
Okay, a straight-up build took 7 minutes without the static fonts. To put it in perspective, Terminal takes 18-30 minutes to build :smile:
Since this isn't part of the CI, or since we can do a stripped down version of this for CI, I'm not worried about the build times at all.

tuananh:
can't wait for the release <3

aaronbell:
@DHowett You should try building *with* the static fonts. :D

DHowett:
image


Here it is in our build pipeline with static + PL + mono :smile:

It's pretty quick, all things considered!

tuananh:
@DHowett where can I download the new build? There's no new release yet :(

madig:
The Python toolchain is a damn snail (did you know that building the Google Noto fonts sequentially takes >24h?). I generate statics in parallel: https://gitlab.gnome.org/GNOME/cantarell-fonts/-/blob/master/scripts/make-static-fonts.py -- warning, will eat all your ram on large amounts of instances. Also, all variable fonts can probably be generated in parallel with multiprocessing along the same lines.

vuongggggg:
Hi, thanks for big effort in this improvement.

How `Semilight` works on Windows, please? I've tried in VSCode with `Cascadia Code SemiLight` `Cascadia Code Regular SemiLight` `Cascadia Code Semilight` but they didn't work. So confused because `Cascadia Code ExtraLight` and `Cascadia Code Light` actually work, SemiLight seems not similar as format of name?

On Mac, `CascadiaCode-Regular_SemiLight` works fine, but can't use it for Windows.

aaronbell:
Oh that's interesting. Normally I'd suggest using VSCode's `Font Weight ` setting, but I note that VSCode doesn't offer the ability to pick a font weight besides on the 100s (similar to how CSS is limited).

The issue here is that while the [Opentype Spec](https://docs.microsoft.com/en-us/typography/opentype/spec/os2#usweightclass) lists 'commonly used' weights only at the 100s, any value is actually allowable. Unfortunately, both CSS and (it appears) VS Code limit the font weight options to the whole 100s following the 'commonly used' values. However, as SemiLight was left off of that original list, the common convention is to insert it at 350. This includes major Microsoft fonts such as Segoe UI Semilight. 

As such, I unfortunately think this is an issue that VS Code will need to resolve, so I've [filed a bug ](https://github.com/microsoft/vscode/issues/103565)on their github. @DHowett, if there's someone you can talk to through your channels to push this above the line, that'd be great :)

aaronbell:
BTW, it appears that there is an existing bug on this issue: https://github.com/microsoft/vscode/issues/101467. I'd suggest heading over there and lending support to it so it can get more attention :). 

This is definitely an issue that VSCode will need to implement.

vuongggggg:
Thank you @aaronbell for giving me more information in this issue.

IllusionMH:
FYI numeric values are supported for `editor.fontWeight` in latest Insiders (`"editor.fontWeight": 350` is now valid, `"350"` as string will be ignored).
`terminal.integrated.fontWeight` is still in progress because it has external dependency on Xterm.js (`"terminal.integrated.fontWeight": "350"` (as string) currently shows warning but works as expected)

PR #305, opened 2020-06-29
Update font to 2007.01, update FONTLOG

DHowett:


mfocko:
Hi, do you plan to release later? Since you've updated the version, but there's no tag nor release on github

PR #308, opened 2020-06-30
Fix casing in build paths for feature files

madig:

Make some paths in build.py case-exact because Linux filesystems tend to be case-sensitive.

## PR Checklist
* [ ] Closes #xxx
* [ ] I've discussed this with core contributors already. If not checked, I'm ready to accept this work might be rejected in favor of a different grand plan.

## Validation Steps Performed

Local build.

### DIFF
--- a/build.py
+++ b/build.py
@@ -204,7 +204,7 @@ if __name__ == "__main__":
                     generator,
                     instance_descriptor,
                     step_set_font_name("Cascadia Code PL"),
-                    step_set_feature_file(INPUT_DIR / "features" / "features_code_pl.fea"),
+                    step_set_feature_file(INPUT_DIR / "features" / "features_code_PL.fea"),
                     step_merge_pl,
                 )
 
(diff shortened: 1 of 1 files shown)


madig:
Me thinks a GitHub Actions thing might be interesting? Or maybe not because the work happens so infrequently?

madig:
DUH. I forgot to build the statics why blow up of course.

madig:
@DHowett what happened to the GitHub Actions thing? Did someone turn it off for PRs? I do see jobs on https://github.com/microsoft/cascadia-code/actions though?!

DHowett:
That's a very good question. We may have simply misconfigured it :smile:

It looks like the action is set to run on `push`; should we expand that?

madig:
It seems to work for PRs now, thanks. Weird, I copy-pasted that `on:` directive from the GHA examples. Oh well.

PR #309, opened 2020-06-30
Generate Variable OTFs

madig:

Generates variable OTFs alongside the variable TTFs. Also, adds compression for all OTFs and fixes a last build path casing issue.

## PR Checklist
* [ ] Closes #xxx

1. Generate variable OTFs for all selected steps. Because why not.
2. Run https://github.com/adobe-type-tools/cffsubr/ on _all_ OTFs to compress them, but only after they have been autohinted (which would undo the compression). cffsubr is a Python-packaged version of the compression (subroutinization) code from Adobe's AFDKO.
3. Fix the last casing feature file path issue preventing static builds on Linux.

## Validation Steps Performed

`python build -S`


requirements.txt
```diff
@@ -2,29 +2,28 @@
...
+compreffor==0.5.0         # via ufo2ft
+cu2qu==1.6.7              # via fontmake, ufo2ft
+fontmake==2.2.0           # via -r requirements.in
+fontmath==0.6.0           # via fontmake
+fonttools[lxml,ufo,unicode]==4.12.1  # via booleanoperations, cffsubr, compreffor, cu2qu, fontmake, fontmath, glyphslib, ufo2ft, ufolib2, vttlib
 fs==2.4.11                # via fonttools
-glyphslib==5.0.1          # via fontmake
-lxml==4.4.1               # via fonttools
-mutatormath==2.1.2        # via fontmake
-psautohint==2.0.1         # for OTF generation
```
DHowett, line 21: so, we lost this line -- should we move psautohint to requirements.in?

DHowett:
YES! Closing and reopened triggered the action.

Sorry if that was a surprise ;P

aaronbell:
I’ve no concerns with this PR, but I’m doing a little investigation about VS that I’d like to get a bit more clarity on before we reissue the release, if that’s OK.

DHowett:
Absolutely.

DHowett:
I'll merge this and wait for your all-clear on a reissue :)

DHowett:
@madig for my own edification, is the major difference between OTF and TTF the preferential use of CFF? Apart from our hinting, that is? As a user in a mixed-platform environment, why should I prefer one over the other?

(No need to answer if it's long/annoying/complicated/there is a document that'll tell me that stuff/etc. :smile:)

DHowett:
Well, this was a surprise after our CI passed it!

/bin/sh: psautohint: command not found
/bin/sh: psautohint: command not found
/bin/sh: psautohint: command not found
/bin/sh: psautohint: command not found
/bin/sh: psautohint: command not found
/bin/sh: psautohint: command not found
/bin/sh: psautohint: command not found
/bin/sh: psautohint: command not found
/bin/sh: psautohint: command not found
/bin/sh: psautohint: command not found
/bin/sh: psautohint: command not found
/bin/sh: psautohint: command not found
/bin/sh: psautohint: command not found
/bin/sh: psautohint: command not found
/bin/sh: psautohint: command not found
/bin/sh: psautohint: command not found
/bin/sh: psautoh

madig:
Argh, yes, psautohint should be in requirements.in. I think the build script didn't error out because subprocess.run does not raise on errors, `check_call` does. I was too asleep yesterday to notice.

madig:
> for my own edification, is the major difference between OTF and TTF the use of CFF? 

Yes, which comes with its own hinting facilities. That hinting happens to display very nicely with FreeType :)

> As a user in a mixed-platform environment, why should I prefer one over the other?

Mh, the CFF2 format (variable OTFs) may not be well supported across applications that implement their own font handling, though the applications that I use have no problem with it. The main difference is probably going to be the visual difference between the hinting. Most Linux distributions that don't change font settings will use fontconfig's slight hinting setting, which translates to FreeType's light hinting, which will make use of the native CFF hinting engine for OTFs, but use the autohinter for TTFs. So by default, on Linux, the TTF hinting you paid for will be ignored and overridden by an autohinter :) Unless you use OTFs :)

The other niche benefit is that Qt applications (but no other, afaik) will use correct text rendering with linear alpha blending and gamma correction when encountering OTFs, but not TTFs, because I haven't found the time and will to completely implement the necessary facilities for TTFs in FreeType.

So, I try to use OTFs exclusively on Linux and TTFs on Windows, because Windows grid-fits OTFs in a weird way.

aaronbell:
BTW, after updating to this version, I observed that the build no longer builds TTFs. Line 144 was converted from saving the variable TTF to only saving the variable OTF file. 

Definitely going to need to fix that :)

madig:
DUH. Sorry, it was just before midnight 😣

madig:
Hm. I just noticed that the variable OTF is glitchy in places. Then I remembered that psautohint stumbles with overlapping glyphs: https://github.com/adobe-type-tools/psautohint/issues/40. Maybe it's too soon to release a variable OTF...

aaronbell:
Yikes! Yes, probably best not to ship it yet. Static is sufficient for now.

PR #320, opened 2020-07-05
Some build.py cleanup

madig:

1. Make static fonts on CI for proper coverage.
2. Move metadata from build.py into the actual sources. Some parameters (e.g. PANOSE) you can set in Glyphs but not a Designspace can be applied anyway with glyphsLib.
3. Disable variable OTF generation due to https://github.com/adobe-type-tools/psautohint/issues/40.
4. Remove a dead function call `step_remove_ligatures` from build.py
5. Simplify static and variable font generation a bit by generating instances once and fully loading a Designspace for variables once. Each production step then gets a copy.

## PR Checklist
* [ ] Closes #xxx

## Validation Steps Performed

CI build.

### DIFF
--- a/build.py
+++ b/build.py
@@ -114,23 +114,26 @@ def build_variable_fonts(designspace, *steps):
 
     print(f"[{familyName}] Done: {file_path}")
 
-    print(f"[{familyName}] Compiling CFF2")
-    file_path_cff2 = (OUTPUT_DIR / file_stem).with_suffix(f".otf")
-    #Do not optimize, because we have to do it again after autohinting.
-    varFontCFF2 = ufo2ft.compileVariableCFF2(designspace,
-       inplace=True,
-       useProductionNames=True,
-       optimizeCFF=ufo2ft.CFFOptimization.NONE,
-    )
+    # XXX: Disable variable OTF (CFF2) compilation until psautohint can better
+    #      deal with overlaps: https://github.com/adobe-type-tools/psautohint/issues/40
+    
+    # print(f"[{familyName}] Compiling CFF2")
+    # file_path_cff2 = (OUTPUT_DIR / file_stem).with_suffix(f".otf")
+    # # Do not optimize, because we have to do it again after autohinting.
+    # varFontCFF2 = ufo2ft.compileVariableCFF2(
+    #    designspace,
+    #    inplace=True,  # Can compile in-place because `designspace` won't be reused here.
+    #    useProductionNames=True,
+    #    optimizeCFF=ufo2ft.CFFOptimization.NONE,
+    # )
 
-    print(f"[{familyName}] Adding STAT table")
-    styleSpace = classes.Stylespace.from_file(INPUT_DIR / "STAT.plist")
-    lib.apply_stylespace_to_variable_font(styleSpace,varFontCFF2,{})
+    # print(f"[{familyName}] Adding STAT table")
+    # lib.apply_stylespace_to_variable_font(styleSpace,varFontCFF2,{})
 
-    print(f"[{familyName}] Saving")
-    varFontCFF2.save(file_path_cff2)
+    # print(f"[{familyName}] Saving")
+    # varFontCFF2.save(file_path_cff2)
 
-    print(f"[{familyName}] Done: {file_path_cff2}")
+    # print(f"[{familyName}] Done: {file_path_cff2}")
 
 if __name__ == "__main__":
     parser = argparse.ArgumentParser(description="build some fonts")
--- a/build.py
+++ b/build.py
@@ -220,7 +220,6 @@ if __name__ == "__main__":
                         generator,
                         instance_descriptor,
                         step_set_font_name("Cascadia Mono NF"),
-                        step_remove_ligatures,
                         step_merge_nf,
                     )
 
@@ -265,7 +264,6 @@ if __name__ == "__main__":
             build_variable_fonts(
                 designspace,
                 step_set_font_name("Cascadia Mono NF"),
-                step_remove_ligatures,
                 step_merge_nf,
             )
 
--- a/build.py
+++ b/build.py
@@ -46,38 +46,6 @@ def step_set_feature_file(n):
 
     return _set
 
-def set_font_metaData(font, sort):
-    font.info.versionMajor = 2007
-    font.info.versionMinor = 1
-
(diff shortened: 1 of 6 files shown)


madig:
Woah. Look at those build times going from 38 to 9 minutes.

build.py
```diff
@@ -46,49 +48,12 @@ def _set(instance):
 
     return _set
 
-def set_font_metaData(font, sort):
```
DHowett, line 49: I believe @aaronbell chose to do this so that we didn't have to do it for a whole heap of different UFO fontinfos, but I'm not certain. It does seem like we're trading DRY for correctness by having it in the build script :smile: which I don't entirely mind

build.py
```diff
@@ -208,99 +163,98 @@ def build_variable_fonts(designspace, *steps):
...
+
         for instance_descriptor in designspace.instances:
+            # Generate instances once.
+            instance = generator.generate_instance(instance_descriptor)
+            glyphsLib.interpolation.apply_instance_data_to_ufo(instance, instance_descriptor, designspace)
 
             build_font_instance(
-                generator, 
-                instance_descriptor,
+                copy.deepcopy(instance),
```
DHowett, line 175: deep copy! of course! this is why I'm not a good Python developer!

build.py
```diff
@@ -208,99 +163,98 @@ def build_variable_fonts(designspace, *steps):
...
                 )
 
                 if not args.no_mono:
                     build_font_instance(
-                        generator,
-                        instance_descriptor,
+                        copy.deepcopy(instance),
                         step_set_font_name("Cascadia Mono NF"),
-                        step_remove_ligatures,
                         step_merge_nf,
```
DHowett, line 213: I think `step_remove_ligatures` was replaced with `step_set_feature_file`. The `Mono` NF version does still need one of those two things, but I'm no longer sure which.

DHowett:
> Look at those build times going from 38 to 9 minutes. 

holy sh-

build.py
aaronbell: Yeah, while I don't mind updating the metadata in the fontinfos, I would prefer a single source of truth in the build process, and that would be in build.py. That way should will always know that those values are set as desired, and nothing will impact them (such as any roundtrips through font editing software).

build.py
aaronbell: Sooooo, I had tried using deepcopy before when setting up this build script and was running into some super irritating issues where the copy wasn't working correctly. I can't recall exactly what the error was, but it took me a week or so of banging my head against the wall to decide to just repeat the generation. I'll check tomorrow if the new build script works without that error (maybe it was due to some other library?) before I feel confident to sign off on this.

build.py
aaronbell: As Dustin mentions, this should be `step_set_feature_file(INPUT_DIR / "features" / "features_mono_PL.fea"),`. I've updated it in my build script, but haven't pushed it out yet :). 

The NF variant would use the same feature file as the PL variant, so the code can by copied directly.

build.py
madig: My thinking was that you're using Glyphs to do the design work anyway, and so you have a single source of truth. GlyphsLib takes care to sync the data across all UFOs. You can also use Glyphs' custom parameters on instances to change stuff.

madig:
> > Look at those build times going from 38 to 9 minutes.
> 
> holy sh-

To be fair, I introduced those times by psautohinting the OTFs. The program is comically inefficient at hinting variable fonts.

build.py
madig: I was indirectly wondering if "Cascadia Mono NF" was still a thing, since the build script would have blown up upon encountering that unknown symbol.

build.py
aaronbell: I would agree with you if we were using GlyphsLib as part of build.py. But since we're not, it seems strange to me to be reliant on converting from Glyphs every time I want to modify the version string. One might say that it is easy enough to go into the files and modify the UFO directly in such a situation, but introduce a manual step when I can automate it via the build file :).

madig: Hm!!! Are you not spending the majority of time in Glyphs? I mean sure, I can reinstate the function with version setting?

build.py
```diff
@@ -208,99 +163,98 @@ def build_variable_fonts(designspace, *steps):
         )
 
     if args.static_fonts == True:
+        # Prepare masters for interpolation.
+        generator = fontmake.instantiator.Instantiator.from_designspace(designspace)
+
         for instance_descriptor in designspace.instances:
+            # Generate instances once.
+            instance = generator.generate_instance(instance_descriptor)
+            glyphsLib.interpolation.apply_instance_data_to_ufo(instance, instance_descriptor, designspace)
```
aaronbell, line 172: I'm not sure I understand the purpose of using glyphsLib here.

madig: It has a function `apply_instance_data_to_ufo` that lets you use various custom parameters Glyphs knows and will apply them to your UFO, e.g. setting PANOSE. Most of them are not known to the Designspace spec, so this is a way to apply them to instances at least.

build.py
aaronbell: Ah. At the moment it isn't being built, no. But I'm hopeful we can get it re-instated.

build.py
aaronbell: Got it :).

build.py
aaronbell: I mean, *I* am. And I do use GlyphsLib to roundtrip. 

But that doesn't mean that I will always be the one building / modifying the font. In a number of cases, Dustin has updated the version string, but he's on a Windows device and it doesn't make sense to force him to do that manually 😄 .

Anyway, I think I'd prefer to keep the Version info, metrics, and GASP table info here. Version info for the reason above. Metrics because I like being able to see them set clearly here (so no matter what happens to the sources in the future, they are locked to what they should be in the official build script), and GASP because I prefer to have VTT-level control over what the GASP table looks like rather than being reliant on Glyphs' implementation.

aaronbell:
I've done some test builds and everything seems to be working fine. So aside from the items I mentioned above, I'm OK with this PR :smile:

build.py
madig: Note: the gasp UFO stuff is the same thing as the actual `gasp` table, so you aren't relying on anything Glyphs here.

PR #348, opened 2020-08-26
Add an Azure DevOps pipeline

DHowett:
This pipeline produces the release archive and makes it so I don't
have to move hundreds of files around on my own every time.

### DIFF
--- a/build.py
+++ b/build.py
@@ -298,7 +298,7 @@ if __name__ == "__main__":
     if args.static_fonts == True:
 
         print ("*** *** *** Autohinting Static Fonts *** *** ***")
-        otfs = list(Path("build").glob("*.otf"))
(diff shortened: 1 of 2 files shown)


build.py
```diff
@@ -297,7 +297,7 @@ def build_variable_fonts(designspace, *steps):
     if args.static_fonts == True:
 
         print ("*** *** *** Autohinting Static Fonts *** *** ***")
-        otfs = list(Path("build").glob("*.otf"))
+        otfs = list(Path("build/static").glob("*.otf"))
```
DHowett, line 300: @aaronbell i think this is correct -- we wanna autohint the statics, not the variables :smile:

aaronbell: Ah yes, that is correct :).

PR #352, opened 2020-08-28
Add OVERLAP flag to simple/compound glyphs, fix interpl. error

### LINKED ISSUE #350: [Linux/FreeType + VariableTTF-specific] Some glyphs (e.g. a '4', or the bottom-right stem of lowercase 'u') look weird at certain sizes, due to mis-handling overlapping paths of variable font. Needs FreeType 2.10.3 and font flag OVERLAP_{SIMPLE, COMPOUND}
# Environment

```none
Cascadia Code version number: 2008.25
Application used to display text: VS Code 1.48.2, Firefox 80.0, Kitty 0.18.3, everywhere
OS platform and version: Ubuntu 20.04.1
Screen resolution (i.e. 220dpi): I don't know
```

# Steps to reproduce

Display a `u` at size 14.7

# Expected behavior

`u` is good-looking

# Actual behavior

`u` looks weird: the bottom right stem has sort of a glitch, where it overlaps to the inside part of the curve:

screenshot_cascadia-code-weird-u_2020-08-27-10:37:25

### LINKED ISSUE #355: -<< looks off with font-weight set to 300
# Environment

```none
Cascadia Code version number:  2008.25
Application (with version) used to display text: PyCharm 2020.2.1, Chrome 85.0.4183.83 
OS platform and version: macOS catalina
Screen resolution (i.e. 220dpi): 192dpi (I think)

Any other software? No
```

# Steps to reproduce

In chrome, change CSS of an element to use: -<<

```css
{
    font-family: 'Cascadia Code';
    font-weight: 300,
}
```

# Expected behavior

Expected the same for >>- but reversed
image

And this is how it looks like with normal fontweight:
image

# Actual behavior

image

PyCharm using this font weight by default and I can't find a setting to change it

aaronbell:

This PR modifies the build script to add "OVERLAP_SIMPLE" to normal glyphs and "OVERLAP_COMPOUND" to component glyphs. 

## PR Checklist
* [x] Closes #350
* [x] Closes #355 - switches contour order to resolve interpolation error at wght 300
* [x] CLA signed.

Under FreeType, "overlapping contours can produce pixels with inflated coverage because FreeType normally integrates the coverage. To mitigate the artefacts, oversampling is necessary (increase resolution 4x4 and average the results)"
([quote from here](https://savannah.nongnu.org/bugs/index.php?59026)). FreeType relies on the "OVERLAP_SIMPLE" and "OVERLAP_COMPOUND" flags to be set to know when this oversampling should take place. 

At current, Cascadia does not contain either of these two flags. This fix will add one or the other to every glyph in Cascadia (depending on type of glyph), thus resolving the issue. Once the ufo2ft library adds functionality to address this issue, the build script modification can be removed. 

The addition of these flags has no impact on the behavior of the font in Mac or Windows environments. 

[Before] Note the bottom right corner of the /u and the middle left of the /four:
91456992-e3cbca00-e851-11ea-8504-3b83c1eab1d7

[After] Displaying properly:
91591579-1e556580-e92b-11ea-9fc4-284c9aec4aaf

Note that this fix will only work on the cutting edge version of FreeType (to be 2.10.3). The current 2.10.2 stable version does not respect the flags. 

For the interpolation error, before:
Screen Shot 2020-09-01 at 9 06 17 AM

After:
Screen Shot 2020-09-01 at 9 05 58 AM

## Validation Steps Performed
Test font reviewed by the bug reporter and found to resolve the problem. 
`--<` rendering checked in FontGoggles


DHowett:
Should we be holding off on this because of the comment here:? https://github.com/googlefonts/ufo2ft/issues/402

aaronbell:
I don’t think so. The only downside to implementing the flag everywhere is that freetype will have slightly slower rendering speed. And if freetype eventually changes their approach to this, we can remove the flag implementation entirely. Or if ufo2ft implements it in a smart way, we can switch to that.

Either way it benefits users now to have the flag set.

PR #361, opened 2020-09-18
Multiprocessing

### LINKED ISSUE #349: Feature Request: Parallel build system
# Description of the new enhancement

Building the fonts for experimentation is currently painfully slow.

# Proposed technical implementation details

1. Use multiprocessing in the build script. Probably doable, I did something for statics generation over at https://gitlab.gnome.org/GNOME/cantarell-fonts/-/blob/master/scripts/make-static-fonts.py
2. Use a proper build system capable of parallel execution. I know of https://pydoit.org/, but find it a bit cumbersome to work with.

I suppose Make would fit the bill but is there a truly cross-platform one one can use without a tedious setup step for at least one major platform? ... Ninja?

aaronbell:
##  Summary of the Pull Request
It builds faster. 

## PR Checklist
* [x] Closes #349 
* [x] CLA signed.

Integrated in Python multiprocessing pools to enable the fonts to build significantly faster. 

On my machine, a full build (all variable & all static) takes 951 seconds. Now it takes **244 seconds**.
Just building the variable fonts has decreased from 64.9 seconds to **29.45 seconds**. 

Much snappier!


aaronbell:
@madig Can you take a look and see if I haven't messed all this up somehow? :)

I'd originally set it up to use multiprocessing.process, but liked how you'd set up yours with the multiprocessing.Pool. Unfortunately, that became problematic when trying to run each variable font build as a separate worker for a couple reasons:
1) no longer can access nested functions (had to convert all of the instantiations to kwargs to bypass)
2) Pool makes all workers daemons by default, which cannot contain another multiprocessing.Pool within them. So I found a handy little script to instantiate a custom version of Pool that has daemon set to False. That way we can run the static instance generator from within each worker. 

Not sure if this is the most efficient way to go about all of this (maybe it would be better to run the static instancer after the variable fonts are built to save processor?), but a ~75% reduction in build time is no joke.

madig:
Cooool :O I'll have a look later. It's true that you have to "flatten" your generation machinery somewhat. I think the instancer can run in parallel to varfont generation, after you read all sources into memory.

I'd still like to pull in a proper build system of sorts, eventually, because hand-reinventing it is a bit ridiculous. There's a ninja Python package that is e.g. used in https://github.com/googlefonts/nanoemoji. Maybe there's an easy-ish way to do this.

aaronbell:
Thanks! Yeah, being able to use something a little more proper than rebuilding would probably be a good thing :). I'll take a look at Ninja, and also found one called "pebble". Seems like there's a lot of options!

madig:
I did a quick profile of the just-VFs case and oof:

grafik

18% of the time is spent deep-copying and restoring the `designspace` object, and then 17% of the time per compilation process.

madig:
I made a cut-down proof of concept build.py for just the variable fonts at https://gist.github.com/madig/03b4be5f5c5ffe5322aae6cc0fb177d2.

Cuts down compilation time from ~25s to ~17s on my machine, compared to the `build.py` in this PR.

grafik

Source loading barely registers now.

madig:
Updated with statics code. The speed gain is more muted here, ~117s vs. ~125s with this PR's build.py on my machine. Can't profile it though because py-spy crashes :(

madig:
Holy shit, I just found the gas pedal. Use pathops instead of the default booleanOperations for statics, cuts down compilation time from 117s to 46s on my machine! Gist updated.

madig:
New py-spy works.

grafik

So! Initializing the `Instantiator` takes a considerable chunk (19% here) vs compiling the statics (31%) and as much as compiling the VFs (17%), so maybe that can be optimized.

aaronbell:
Wow! Thanks for your investigations on this. Definitely learning some neat tricks about code optimization :). 

A question—in looking at your code, you're rebuilding the designspaces for the static instances. Do you think it would be faster to try and retrieve the updated ones from the multiprocessing workers ( build_font_code ) and use those rather than rebuilding?

madig:
Hm, I don't know, but I have a hunch that it won't help much, if at all. For once, you'd have to coordinate the tossing over somehow (with Queues maybe) and then the transfer has to not take too long. The pickled (serialized to bytes to send over the wire) Instantiator object is ~3MB and then you have to reconstruct it. Not sure how efficient Python is at doing all that... Python 3.8 comes with https://docs.python.org/3/library/multiprocessing.shared_memory.html, maybe that can be used? I'm a bit of a multiprocessing noob and also not sure whom to ask.

aaronbell:
Thanks again for all your work on this! It is definitely building faster now 👍 

I did a little rework on your last version to organize the code and reduce the amount of repetition (now under 300 lines of code!). From what I can tell, it appears to build at the same speed as your version: https://gist.github.com/aaronbell/98e92bfe7fb08be02395390f1acd9438

>  I'm a bit of a multiprocessing noob and also not sure whom to ask.

You're definitely doing better than me! 😆 I tried using py-spy and had a hard time figuring out what the heck it is trying to tell me. 

> PS: I think you may want to make a topic branch on your repo fork and open a PR from that? This PR is pulling in 30 commits in total, many of them unrelated?

I thought it is because they haven't rebased the git repository in a long time (so it is pulling all of my PR from this branch forever). Would taking the branch approach work better?

madig:
> I did a little rework on your last version to organize the code and reduce the amount of repetition

Cool, updated my gist with yours plus formatting and filling in all typing. It helps code intellisense with IDEs. Runs in the same time, yes. In the preparation function, you might want to turn the `else:` into a catch-all that raises an exception, in case you mistype something and it silently goes the "Code" path.

> I tried using py-spy and had a hard time figuring out what the heck it is trying to tell me.

It's hard to read when you have multiprocessing going on, but do have a look at https://github.com/jlfwong/speedscope. It explains some of what you see.

> I thought it is because they haven't rebased the git repository in a long time (so it is pulling all of my PR from this branch forever). Would taking the branch approach work better?

Mh, it's very easy to make a mess with Git, so I always work in topic-branches. If a PR starts to show more commits than what I put in, I get nervous.

aaronbell:
@madig, btw, I installed Pylance and it looks great, but I'm getting this strange problem from it: 
Screen Shot 2020-09-21 at 9 08 15 AM

Clearly the script is working fine since it builds without issue, but Pylance doesn't seem to like that particular line. Are you seeing something like that too?

madig:
@aaronbell Hm, I'm not seeing that one? Are you inside the venv where the lib is installed?

aaronbell:
@madig I tend to run the file from the cascadia-code folder rather than the venv folder. Turns out that Pylance assumes that the open VSCode folder is the root and searches from there. So I had to add the proper venv lib as an "extra" path to search and then it went away. Confusing!

madig:
venv management is annoying, yes. You need to open the project root in VSCode and activate the venv ("select interpreter" or something, should find the venv automatically) and then Pylance should pick it up.

aaronbell:
Yeah, I just used this (in Pylance's preferences) to point it at the venv: 
Screen Shot 2020-09-21 at 3 08 11 PM

madig:
Hope you kept that setting in the workspace or folder section instead of the global one :D

PR #363, opened 2020-09-21
Multiprocessing v2

### LINKED ISSUE #349: Feature Request: Parallel build system

aaronbell:
It builds faster.

## PR Checklist
[x] Closes #349
[x] CLA signed.

Integrated in Python multiprocessing pools to enable the fonts to build significantly faster. Also integrates skia-pathops for faster static font generation, and rejigs the build process to get rid of the really slow copy commands. Thanks @madig!!!

See discussion here: #361 for more details about what's happened and why. :)

Much snappier!


DHowett:
Just gotta cover my bases--@madig: you're comfortable with the terms of our CLA (which I believe you've already signed!) applying to your contributions here and that they'll fall under the license that the work in this repo is already under?

I know you contributed the initial prototypes for the build scripts, and a bunch of other stuff, but it doesn't hurt to check especially when your name isn't attached to this PR :smile:

madig:
Yes.

DHowett:
Thanks. 🥂

.gitignore
```diff
@@ -336,4 +336,6 @@ newVTT.xml
 testground.py
 buildTest.py
 converter.py
-salishan/
\ No newline at end of file
+salishan/
+fixTSI1.py
+sources/vtt_data/vtt_build/from fontTools.ttLib import TTFont.py
```
DHowett, line 341: this last line seems weird to me :smile:

aaronbell: Shhhhh. I put it in gitignore for a reason! 

I'm somewhat miffed y'all aren't interested in the 'fixTSI1.py' I added too :P

DHowett: I'm definitely interested in that one! I just don't know enough to know what a TSI1 is

(I've since looked and learned some more stuff about VTT!)

madig: TSI1 is where VTT puts the assembly stuff compiled from TSI3 (think CVT template, the glyphs' VTT Talk, ...).

build.py
```diff
@@ -62,95 +62,175 @@ def set_font_metaData(font, sort):
...
+            step_set_feature_file(FEATURES_DIR / "features_mono.fea", source.font)
+            step_set_font_name(name, source.font)
+        elif "PL" in name:
+            step_set_feature_file(FEATURES_DIR / "features_code_PL.fea", source.font)
+            print(f"[{name} {source.styleName}] Merging PL glyphs")
+            step_merge_glyphs_from_ufo(
+                NERDFONTS_DIR / "NerdfontsPL-Regular.ufo", source.font
+            )
+            step_set_font_name(name, source.font)
+        else:
```
madig, line 108: You might want to make this an explicit

```
if name == "Cascadia Code":
    ...
elif ...
else:
    raise ValueError(f"Unknown font configuration: {name}")
```

Because this will catch typos.

requirements.txt
```diff
@@ -22,6 +22,7 @@ pyclipper==1.1.0.post3    # via booleanoperations
 pyparsing==2.4.7          # via vttlib
 pytz==2020.1              # via fs
 six==1.15.0               # via fs
+skia-pathops==0.4.1
```
madig, line 25: Add this in requirements.in without the version pin and do `pip-compile requirements.in`.

.gitignore
aaronbell: TSI1 also happens to be where all of the component-related references are stored and mark positioning-related tweaks are made.

build.py
aaronbell: Ah yeah. I forgot to do that :x

requirements.txt
aaronbell: Updated

build.py
aaronbell: done

.gitignore
aaronbell: (I'll put all of my test scripts in a little folder if it'll make y'all feel better. )

DHowett: I don't mind them! I was more worried about the line that looked like some python code that had accidentally made its way in. 😁

DHowett:
Alright, i believe I understand how this works. Very cool!

build.py
```diff
@@ -1,53 +1,53 @@
...
+import subprocess
+from pathlib import Path
+from typing import cast
 
+import cffsubr.__main__
 import fontmake.instantiator
 import fontTools.designspaceLib
+import fontTools.ttLib
+import fontTools.ttLib.tables._g_l_y_f as _g_l_y_f
+import psautohint.__main__
```
DHowett, line 14: it's not concerning to call `__main__` from another _command_ module, right? What if it has global state? I've never seen this pattern ("invoke a command's implementation directly when it wasn't designed as a library") before...

madig: I hope that it does not have relevant global state but haven't checked. I like to avoid launching subprocesses because it's so slow on Windows.

DHowett (approved):

DHowett:
on our build farm...

### old

image

### new

image


_holy crap_.

aaronbell:
tenor

madig:
Just saw the Azure pipeline thing. Why are there so many steps in there? The VTT code can be "compiled" with https://pypi.org/project/vttLib/ (actually it takes the TSI1 code produced by VTT from TSI3 code and present in the TTX dump and translates it into something fontTools understands). I actually missed that when going over the build script. The WOFFing can be done in the same way the other stuff is done with pools (the compression is very CPU heavy), the script currently does it serially. So there's more scope for streamlining. Also, if you have Azure, I suppose you don't need the GitHub workflow?

DHowett:
Ah, I didn't know that vttlib could compile VTT. Not being aware of that, I asked the owners of VTT if I could produce a NuGet package of their CLI tool and ship it as a dependency here. :smile:

The pipeline is relatively new, in that it postdates the CI workflow, but I don't see the harm in having both. This one produces final release artifacts whereas the PR/CI workflow just makes sure things aren't instantly horked.

DHowett:
So with vttLib we could do this all in one shot? Compile and strip the vtt source and effectively build the font for shipping all from python?

madig:
Yep. PR incoming.

aaronbell:
Please set it up with an args so I can export a VTT compatible version 🙏

PR #365, opened 2020-09-21
Streamline the build by shipping VTT w/ vttlib+WOFF2 at same time

madig:

1. Make WOFF2 fonts from the same build script, in parallel.
2. Compile VTT code by default for shipping (turn off with `-V` as requested).

The command line to make all files for a release now is `python build.py -S -W`.

 
## References

See https://github.com/microsoft/cascadia-code/pull/363 for comments.

## PR Checklist
* [x] Closes an inefficiency
* [x] I've discussed this with core contributors already. If not checked, I'm ready to accept this work might be rejected in favor of a different grand plan. Issue number where discussion took place: #xxx

1. The current Azure pipeline is a bit all over the place. WOFF2 fonts can be geenrated from build.py directly, whereas the separate script does it serially. Do note that keeping the VTT source in (`-V`) does not trip the WOFF2 generator function now.
2. No need for running VTT, vttLib can compile VTT code as well (minus some small stuff that I don't think is relevant here)

## Validation Steps Performed

Running stuff manually.


madig:
Oh by the way, the font no longer needs to be built on macOS by default. Remember how that had to happen because the overlap removal lib behaved differently depending on the platform? That wouldn't have been a problem, were it not for the manual VTT hinting code that is breaks if point indices shuffle around. Now, the VTT code is on the VF, where overlaps are not removed. Statics are autohinted, so this is no longer an issue!

madig:
Also also: don't you want to hook in Azure into GitHub's PR checks thingy? Do you want to keep the GitHub action?

DHowett:
Since it no longer takes 17 minutes to do a full run, yeah, I'd love to streamline the font generation into a single phase and wire it up to pull request CI.

I'm planning to handle that separately, because I need to ask for a dev.azure.com/ms project to do cascadia things (or just tie it to terminal ;P)

build.py
```diff
@@ -155,10 +172,13 @@ def compile_variable_and_save(
 
     print(f"[{familyName}] Merging VTT")
     vttLib.transfer.merge_from_file(varFont, VTT_DATA_FILE)
+    if vtt_compile:
```
DHowett, line 175: mind adding in a log for Compiling VTT?

```suggestion
    if vtt_compile:
        print(f"[{familyName}] Compiling VTT")
```

DHowett: Eventually I'll get up the spare time to work a logging context into the multiprocessing steps so we can track time taken for each one ;P

DHowett:
Curiously, there's some differences in the output files between vttshell and vttlib.

* New output files are ~200kb larger (looks like just a difference in compression?)
* ttx shows...
   * new output has a different flags (**OLD** `00000000 00001011` **NEW** `00000000 00000011`)
   * `PUSHx A B` became `PUSHx A` `PUSHx B` (this one's just interesting, not a concern)
   * a number of points now state `overlap=1`
   * a number of components now have the right flags (`| 0x400`)

_oh no_. 😀 

It legitimately looks like 2009.14 _and 2009.21_ never had the fix for #350 in the final output. I wonder if VTT was stripping it?

font build|count of flags=0x404/0x604|
-|-
this pr|521
2009.21|0
2009.14|0
2007.15|0

@aaronbell any ideas?

DHowett:
I guess my only worry is the file size then!

madig:
Ah, the joys of not having unit tests for fonts! :grin: 

The size diff is curious indeed.

fontTools handles PUSHes differently, so I'd expect that.

aaronbell:
I can't imagine why VTT would strip that flag. :x Wellllllll that's interesting.

madig:
Which output files are you looking at?

madig:
(I once worked around a bug where a spacing glyph had nothing but a comment in its TSI1 code section and that caused VTT to insert a lone point into the glyph that tripped FontValidator)

aaronbell:
It appears that the "ship" command from VTT strips that flag, or I should say, it changes the flags to whatever it wants. I just took a look at the version I produced for the Salishan extensions and saw the difference.

madig:
Mh. The newest VTT 6.33 can corrupt VFs under certain circumstances, so this does not surprise me. Also, the flag is only needed on non-MS platforms, so...

sources/vtt_data/CascadiaCode.ttx
```diff
@@ -8708,7 +8708,7 @@ IUP[X]
 /* VTTTalk Unicode 0x2202 */
 /* ACT generated Mon Sep 14 09:57:28 2020 */
 
-SOFFSET[R], 624, 1202, 0, -1, 0, 0, 1
+OFFSET[R], 624, 1202, 0
```
aaronbell, line 8711: What's the cause of this change?

madig: vttLib does not know what SOFFSET is (I don't either... scaled offset?) but it seems useless so I removed the superfluous bits.

aaronbell: Ah, you know, it looks like I had used a mirrored component there. So it is an offset with a transform applied. 
I should get rid of that in the source and add some hints. Will do that (later).

madig: Oooh wait, does this flip something? :open_mouth:

DHowett:
@madig as for which files, I was comparing CascadiaCode.ttf from the latest release package (2009.21) and the one produced from this PR with VTT compilation _enabled_.

I’m getting the feeling I should recall 2009.21 and ship an update in short order that supersedes it, with apologies to the community, because we never got the overlap flags. :)

sources/vtt_data/CascadiaCode.ttx
DHowett: So is this line valid still?

aaronbell: @DHowett I don't have the time right now to make the fix. Will do so tomorrow morning. 
Because the glyph will change from a component one to an outline one, all of that will change.

madig:
```
(venv) ~/E/cascadia-code (streamlining|✔) $ ll build/ttf/
-rw-r--r--. 1 nikolaus nikolaus 370K 21. Sep 23:29 CascadiaCodePL.ttf
-rw-r--r--. 1 nikolaus nikolaus 345K 21. Sep 23:29 CascadiaCode.ttf
-rw-r--r--. 1 nikolaus nikolaus 351K 21. Sep 23:29 CascadiaMonoPL.ttf
-rw-r--r--. 1 nikolaus nikolaus 326K 21. Sep 23:29 CascadiaMono.ttf

(venv) ~/E/cascadia-code (streamlining|✔) $ ll /tmp/CascadiaCode-2009.21/ttf/
-rw-r--r--. 1 nikolaus nikolaus 371K 21. Sep 19:28 CascadiaCodePL.ttf
-rw-r--r--. 1 nikolaus nikolaus 346K 21. Sep 19:28 CascadiaCode.ttf
-rw-r--r--. 1 nikolaus nikolaus 352K 21. Sep 19:28 CascadiaMonoPL.ttf
-rw-r--r--. 1 nikolaus nikolaus 327K 21. Sep 19:28 CascadiaMono.ttf
```

I see files 1kb smaller because pathops seems to be slightly more space-efficient?

DHowett:
I'm a dork. 

image


guess which column

madig:
Mod?

sources/vtt_data/CascadiaCode.ttx
DHowett: Sorry, just so we're on the same page: I should not ship a font built out of this script because this would damage one of the glyphs?
Sorry if that's a dense question :smile:

madig:
Also, don't do a quick fix release until @aaronbell fixes the SOFFSET thing I found :grin:

sources/vtt_data/CascadiaCode.ttx
madig: Potentially. Never understood what VTT needs these OFFSET things for. I'll have a quick look with FreeType.

aaronbell: I think you can actually just remove the OFFSET entirely and then ship it. That way it isn't being modified at all by hinting instructions and I can correct it when I have a chance. But I'd suggest checking 0x2022.

madig: So that particular line belongs to `partialdiff`, glyph ID 859 in CascadiaCode.ttf. FreeType shows no immediately visible diff between old and new build...

madig: (I must now sleep. One can always re-release another day!)

DHowett:
Thanks everyone for digging in on this :smile:

sources/vtt_data/CascadiaCode.ttx
DHowett: I decided to hold this even though we're timing for a Terminal release -- no point in rushing it, and Terminal only has to work on Windows whereas the general-use build of this font has to work right everywhere :smile:

DHowett:
I also appreciate you getting rid of my "move heaps of files around for no reason" pipeline by just having the build output go into the right places ;P

sources/vtt_data/CascadiaCode.ttx
aaronbell: Thanks @DHowett. Turns out I'll be able to get to this tonight, if that helps since we have such a blazing fast build process now ;). Or just wait until tomorrow because I'm sure you have better things to do. Haha.

DHowett: Eh, I already pressed the Terminal builds and it'll be nicer for us to release "2009.22" on Sep 22.

If you want to propose a change tot his file in another branch and @madig can merge master, I'm all for it.

Thanks :D

aaronbell:
@madig I'm a little concerned by the number of changes in the `fpgm` table w/r/t pushes. Do you have anything I can look at that discusses how fontTools handles theses differently? In my ttx roundtrips I've never seen this sort of variance.

aaronbell:
OK, I submitted a new push to this PR with a corrected partialdiff and quotereversed glyph. This removes the reversed components and adds hinting. 

HOWEVER, at current the implementation of vttLib cannot be used for Cascadia Code. It appears not to support the generation of the CVAR table, which is necessary for varying the CVT values across the variable axis. 😭

DHowett:
Augh i forgot to mention CVAR in my analysis!

VTT corrupts VF fonts, and vttLib can’t do CVAR? It sounds like our hands are tied...?

DHowett:
I wish we had stripped dates out of the VTT export the first time; they cause thousands of lines of diffs that don’t amount to much!

aaronbell:
Yup. My only consolation is that it won't impact Windows users at all.

As for the dates in the VTT export, VTT will automatically generate those every time I save the font. Hm. But I can write a script to remove them. Will just have to run it every time I update the VTT source.

aaronbell:
Alright, also removed all of the comments about date generation and how things are generated from the TSI1 and TSI3 tables. There's been some other formatting changes thanks to how Elementtree deals with XML, but moving forward these things should be cleaner.

moyogo:
> HOWEVER, at current the implementation of vttLib cannot be used for Cascadia Code. It appears not to support the generation of the CVAR table, which is necessary for varying the CVT values across the variable axis. 😭

You can use vttLib to generate interpolatable masters and use fontTools.varLib to generate a variable font with CVAR, unless the CVAR is not straightforward.

madig:
If I remember correctly, `cvar` is supported in that you have to merge the VTT TTX dump into the VF, compile everything in VTT (thus generating a cvar) and dump it again with vttLib. The newest vttLibs should dump `cvar` along with the rest: https://github.com/daltonmaag/vttLib/blob/v0.9.1.post2/src/vttLib/transfer.py#L14-L50. It can also be done manually, but requires understanding how the hell varLib generates this stuff: https://github.com/fonttools/fonttools/blob/4.15.0/Lib/fontTools/varLib/__init__.py#L336-L364.

@aaronbell: I have no info on the PUSH implementation or how it might affect things. I think the only real answer can be given by building on e.g. the examples in https://github.com/rougier/freetype-py/tree/master/examples and doing something like

```
font_orig = freetype.Face("old/CascadiaCode.ttf")
font_new = freetype.Face("new/CascadiaCode.ttf")

for size in range(8, 64):
    for glyph_name in font_orig:
        for wght in range(100, 900, 100):
            ... set both fonts to current wght ...
            ... render both with hinting ...
            ... assert both have same points ...
```

aaronbell:
Well, for the sake of getting an update to this font out it sounds like the easiest solution is to do some manual tweaks to the font (adjusting the head table, merging the correct cvar, etc) at the end of the production script.

aaronbell:
I've pushed an update that imports the previously-made CVAR table, and modifies the flag. So it should be good now. 

I think it would still be interesting to better understand the differences in rendering compiling, but I'm OK with getting this update in as is to fix the current problem now, and I can evaluate the rendering at a later time.

DHowett:
@madig sorry for the both of us partying all over your branch -- I know that it can be disruptive and loud! However, your vtt diffs are down to only the actual code changes :smile:

DHowett:
Is it possible to have the cvar table live in the existing vtt data ttx?

madig:
No problem! Still at work so I cannot legally contribute right now.

I can however point out that cvar can live in the TTX dump. Just dump the TTX with vttLib:

```
> python -m vttLib dumpfile
usage: python -m vttLib dumpfile [-h] [-v | -q] INPUT.ttf [OUTPUT.ttx]
python -m vttLib dumpfile: error: the following arguments are required: INPUT.ttf
```

madig:
```python
import freetype

font_old = freetype.Face("/tmp/CascadiaCode-2009.21/ttf/CascadiaCode.ttf")
font_new = freetype.Face(
    "/home/nikolaus/Entwicklung/cascadia-code/build/ttf/CascadiaCode.ttf"
)

for size in range(8, 65):
    font_old.set_char_size(size * 64)
    font_new.set_char_size(size * 64)
    for wght in range(200, 800, 100):
        font_old.set_var_design_coords((wght,))
        font_new.set_var_design_coords((wght,))
        for unicode_value, glyph_id in font_old.get_chars():
            font_old.load_char(unicode_value)
            font_new.load_char(unicode_value)

            if font_old.glyph.outline.points != font_new.glyph.outline.points:
                print(hex(unicode_value), glyph_id)
```

```
(venv) /tmp $ python a.py | sort | uniq
0x201b 720
0x2202 859
```

So only the changed glyphs are different according to FreeType.

aaronbell:
Ship it!

madig:
I can't!

DHowett:
Excellent!

In that case, I'll see about merging the cvar table into the ttx (re-exporting it stripped comments in a different way, OF COURSE, so the diff is AS ALWAYS huge. @madig how do you all handle that? is hinting program code simply unreviewable?).

It looks like the new vttLib also exports `cv`. Should we care about that/carry it?

DHowett:
@aaronbell the integral ppem flag (bit 3 in head.flags) -- it's not been on for any of our prior builds. What's it needed for for our font?

madig:
Yes, we always use vttLib and never export fonts from VTT directly, then we roundtrip the code between VTT and vttLib a few times until diffs are zero and pin versions so stuff changes as little as possible. Reviewing code is humanly impossible (I strongly dislike TT asm and its stack-ness) so we do this from the start and people look at stuff manually as they go. In case of huge diffs, they close their eyes, pray and click "commit".

DHowett:
In that case, I'll make sure it lands in the final ttx ;P

aaronbell:
@Dhowett huh? You mentioned it in your review above:
> new output has a different flags (OLD 00000000 00001011 NEW 00000000 00000011). 
So the modification I made was just to restore it back to what it should be.

DHowett:
okay, w/ minimal changes cvar is in the ttx dump, i kept the ppem integer flag (???), etc

DHowett:
Oh, you know what. I was comparing the wrong thing. Yes, okay, we're good. Ignore that question. LOL

aaronbell:
I'm super confused by the need for the `cvt` table. Does that directly influence the cvt code in the font? As in, do I need to manage that as well if I need to add new cvt information, or can it just be a stub and export will read (as it should) from the TSI tables correctly?

DHowett:
If we don't need to keep `cvt` in this xml document, is having cvar in there acceptable? :smile:

aaronbell:
Well, apparently we need the `cvt` there for vttLib to be able to import `cvar`, which I hadn't realized. If it is so crucial, I'd prefer to just keep it separate and import like I'd done before.

DHowett:
Fair enough. :)

DHowett:
Back to what it was before, and I'm happy to sign off if you are. :smile:

aaronbell:
Yup!

DHowett:
Outstanding. Thanks @madig, and @aaronbell sorry for stepping on your toes a little bit. My sense for "cleanliness" took over, and I was worried about having more "manual" phases. At least they can all be described in Pythonand we have one gold master build script. :smile:

madig:
The cvt table in the TTX dump is ignored on import. It's really only there because fonttools' cvar table importer really wants to see a cvt table first or else.

aaronbell:
Good to know! So essentially one could put a stub CVT table in there and call it good.

madig:
I think! I forgot the details. Too many other battles to fight.

PR #366, opened 2020-09-22
Strip date comments from CascadiaCode.ttx's VTT data

DHowett:
@aaronbell this is my proposal for stripping comments -- it's less intrusive than the changes ElementTree made, and it's a quick script any other contributors can use.

To use,

```
./stripVttComments.sh
```

That should be it.

If you agree with this approach, I can handle merging master into Nikolaus' branch to reduce the churn over there :smile:


DHowett:
also, I see the value in the other comments that VTT emits; somebody who wants to look at the source and _actually understand what it's doing_ might want to keep them!

DHowett:
(I'm afraid of the 13k line diff in the ttx in #365, so I wanted to pull it out and simplify it!)

aaronbell:
It seems alright to me (and better to keep the glyph name comments), though I'd suggest also removing:
`/* GUI generated Sun Jun 28 08:35:23 2020 */`
`/* ACT generated Mon Sep 14 09:57:28 2020 */`

DHowett:
Oh, good call!

DHowett:
Got em!

PR #367, opened 2020-09-22
Azure Pipelines: add PR trigger, streamline to one phase

DHowett:
This PR introduces a PR trigger to the azure build pipeline, now
that it only takes about 3 minutes end-to-end. Every build will now
fully ship the font in all variants and produce a font archive.

It also removes the GitHub actions workflow.


DHowett:
This actually worked! The build even produced a font zip!

DHowett:
i'll try one more thing ... a Windows build agent

DHowett:
ttfautohint is going to be a pain to get on Windows; I'll defer that change for now.

madig:
I'd be interested in a performance comparison between all platforms. Windows is usually the slowest by far in our infrastructure (disk I/O is atrociously slow). You can cache the windows binaries for ttfautohint maybe or update and use https://github.com/fonttools/ttfautohint-py.

madig:
Mh, I need to login to Azure to see details of the build job? Is it not integrated with GH's Checks tab?

DHowett:
@madig so, Azure Pipelines has an N-1 problem here. I set up PR runs with PR triggers in the same tenant that hosts Windows Terminal’s builds, but because there hasn’t been a revision on master that has the PR trigger set up it’s not being kicked off for PRs.

I ran our private copy of the pipeline against it to validate before this merges, which will hopefully be the last time I need to do that. It’s private only in that it isn’t visible externally-it uses the same YAML config :)

DHowett:
@madig [here's one from the visible tenant](https://dev.azure.com/ms/terminal/_build/results?buildId=111508&view=results) :smile:

madig:
(You can still switch from macOS to Linux for building, we don't need macOS anymore)

DHowett:
Small steps :smile:

PR #370, opened 2020-09-27
Set the names of the static instances properly

### LINKED ISSUE #369: according to fontconfig, familyname no longer contains style name :(
It looks like one of our recent build changes broke family names in the static TTFs/OTFs:

### cascadia code 2008.xx

Family name _for the most part_ contains the style name (in addition to just "Cascadia Code")

```
% for i in CascadiaCode-*; do echo "$i"; fc-query "$i" | grep family:; done 
CascadiaCode-Bold.otf
	family: "Cascadia Code"(s)
CascadiaCode-ExtraLight.otf
	family: "Cascadia Code"(s) "Cascadia Code ExtraLight"(s)
CascadiaCode-Light.otf
	family: "Cascadia Code"(s) "Cascadia Code Light"(s)
CascadiaCode-Regular.otf
	family: "Cascadia Code"(s)
CascadiaCode-SemiBold.otf
	family: "Cascadia Code"(s) "Cascadia Code SemiBold"(s)
CascadiaCode-SemiLight.otf
	family: "Cascadia Code"(s) "Cascadia Code SemiLight"(s)
```

### cascadia code 2009.22

Family name no longer contains style

```
% for i in CascadiaCode-*; do echo "$i"; fc-query "$i" | grep family:; done                      
CascadiaCode-Bold.otf                                                                                    
	family: "Cascadia Code"(s)
CascadiaCode-ExtraLight.otf
	family: "Cascadia Code"(s)
CascadiaCode-Light.otf
	family: "Cascadia Code"(s)
CascadiaCode-Regular.otf
	family: "Cascadia Code"(s)
CascadiaCode-SemiBold.otf
	family: "Cascadia Code"(s)
CascadiaCode-SemiLight.otf
	family: "Cascadia Code"(s)
```

### (for reference): Fira Code 5.2

```
FiraCode-Bold.ttf
	family: "Fira Code"(s)
FiraCode-Light.ttf
	family: "Fira Code"(s) "Fira Code Light"(s)
FiraCode-Medium.ttf
	family: "Fira Code"(s) "Fira Code Medium"(s)
FiraCode-Regular.ttf
	family: "Fira Code"(s)
FiraCode-Retina.ttf
	family: "Fira Code"(s) "Fira Code Retina"(s)
FiraCode-SemiBold.ttf
	family: "Fira Code"(s) "Fira Code SemiBold"(s)
```

I noticed this when I upgraded Cascadia on my Linux machine and all of my applications stopped being able to find it. I have yet to figure out how to configure them to locate Cascadia based on the style field.

aaronbell:

Now builds static instances correctly. 

## PR Checklist
* [x] Closes #369
* [x] CLA signed.

As far as I can tell, what essentially happened here is that _instances_ in the designspace were not being renamed, even though the sources were being renamed. As a result, the variable fonts would build correctly, but the static instances were not. The font naming approach has now been changed and is producing reliable results. 

What's particularly vexing is that for the life of me I can't figure out how the code actually worked before, because as far as I can tell, we've _never_ modified the instance code before.

## Validation Steps Performed
Checked tables in OT Master

### DIFF
--- a/build.py
+++ b/build.py
@@ -218,7 +218,7 @@ def compile_static_and_save(instance: ufoLib2.Font, name:str) -> None:
         optimizeCFF=ufo2ft.CFFOptimization.NONE,
     )
 
-    file_name = name.replace(" ", "") + "-"  + instance.info.styleName
+    file_name = f"{family_name}-{style_name}".replace(" ", "")
     file_path_static = (OUTPUT_STATIC_TTF_DIR / file_name).with_suffix(".ttf")
     file_path_static_otf = (OUTPUT_STATIC_OTF_DIR / file_name).with_suffix(".otf")
 
--- a/build.py
+++ b/build.py
@@ -39,10 +39,10 @@ NERDFONTS_DIR = INPUT_DIR / "nerdfonts"
 
 
 def step_set_font_name(name: str, source: ufoLib2.Font) -> None:
-    source.info.familyName = name
+    source.info.familyName = source.info.familyName.replace("Cascadia Code", name)
     # We have to change the style map family name because that's what
     # Windows uses to map Bold/Regular/Medium/etc. fonts
-    source.info.styleMapFamilyName = name
+    source.info.styleMapFamilyName = source.info.styleMapFamilyName.replace("Cascadia Code", name)
 
 
 def step_merge_glyphs_from_ufo(path: Path, instance: ufoLib2.Font) -> None:
@@ -122,8 +122,9 @@ def prepare_fonts(
             print("Variant name not identified. Please check.")
         set_font_metaData(source.font)
     for instance in designspace.instances:
-            instance.familyName = name
-            instance.styleMapFamilyName = name
+        instance.name = instance.name.replace("Cascadia Code", name)
+        instance.familyName = instance.familyName.replace("Cascadia Code", name)
+        instance.styleMapFamilyName = instance.styleMapFamilyName.replace("Cascadia Code", name)
 
 
 def to_woff2(source_path: Path, target_path: Path) -> None:
@@ -155,7 +156,7 @@ def build_font_static(
     prepare_fonts(designspace, name)
     generator = fontmake.instantiator.Instantiator.from_designspace(designspace)
     instance = generator.generate_instance(instance_descriptor)
-    compile_static_and_save(instance)
+    compile_static_and_save(instance, name)
 
 
 # Export fonts
@@ -196,8 +197,8 @@ def compile_variable_and_save(
     print(f"[{familyName}] Done: {file_path}")
 
 
-def compile_static_and_save(instance: ufoLib2.Font) -> None:
-    family_name = instance.info.familyName
+def compile_static_and_save(instance: ufoLib2.Font, name:str) -> None:
+    family_name = name
     style_name = instance.info.styleName
     print(f"[{family_name}] Building static instance: {style_name}")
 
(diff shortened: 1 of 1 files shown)


aaronbell:
@madig - I wonder if when you updated the requirements.txt file in our last PR that something changed in the instances from the designspace file. Because we've never modified the instance naming before and it always worked OK.

build.py
```diff
@@ -215,7 +218,7 @@ def compile_static_and_save(instance: ufoLib2.Font) -> None:
         optimizeCFF=ufo2ft.CFFOptimization.NONE,
     )
 
-    file_name = f"{family_name}-{style_name}".replace(" ", "")
+    file_name = name.replace(" ", "") + "-"  + instance.info.styleName
```
DHowett, line 221: I think this line has the same effect as it did before editing, but without the format interpolation it's a little less clear. any reason to change it? :)

madig:
I'll take a look later, gotta run now. If you haven't, try changing the static fonts compilation function to also change the family names for all instances before instantiating the instantiator and then remove the single name setter after it.

build.py
aaronbell: The family name is being set as “CascadiaMonoExtralight” so the filename becomes “CascadiaMonoExtralight-Extralight” whereas we wanted “CascadiaMono-Extralight”. That’s why I changed it.

aaronbell:
> I'll take a look later, gotta run now. If you haven't, try changing the static fonts compilation function to also change the family names for all instances in the designspace before instantiating the instantiator and then remove the single name setter after it.

:)

build.py
aaronbell: (As a side note, if we were build Mac platform naming, we could actually leave the family name ame as “Cascadia Mono” but that doesn’t work for the Microsoft platform. Instead we set the family name to “Cascadia Mono Light” (subfamily “Regular”) and then use the preferred family name fields to set “Cascadia Mono” (subfamily “Light”).

DHowett: Yeah.. wish it were that easy!

I'm still confused -- this code (skipping the bits in between) changes from..

```
file_name = f"{family_name}..."
```

to

```
name = family_name
file_name = name + "..."
```

I think they resolve to the same thing, but the concatenating version takes more mental effort to parse :)

madig: Huh. `f"{family_name}-{style_name}".replace(" ", "")` should stay because now `family_name = name` and `style_name = instance.info.styleName`, so this is a null-change. Also, the existing line despaces everything, style_name could potentially have a space.

madig:
> :)

It's like you read my mind.

build.py
aaronbell: Oh I see. I think I may have confused myself. I’m ok with reverting that line.

aaronbell: done

madig:
> @madig - I wonder if when you updated the requirements.txt file in our last PR that something changed in the way instances are generated from the designspace file. Because we've never modified the instance naming before and it always worked OK.

I think this is a case of setting things in a different order. Font naming is annoying as is and maybe the Designspace ergonomics aren't the best here... you could probably get rid of the `familyname`, `stylemapfamilyname` and `stylemapstylename` instance fields and ufo2ft would do thee right thing for you, as long as you set the correct `familyName` (and unset `styleMapFamilyName`) in the sources.

aaronbell:
> > @madig - I wonder if when you updated the requirements.txt file in our last PR that something changed in the way instances are generated from the designspace file. Because we've never modified the instance naming before and it always worked OK.
> 
> I think this is a case of setting things in a different order. Font naming is annoying as is and maybe the Designspace ergonomics aren't the best here... you could probably get rid of the `familyname`, `stylemapfamilyname` and `stylemapstylename` instance fields and ufo2ft would do the right thing for you, as long as you set the correct `familyName` (and unset `styleMapFamilyName`) in the sources.

Right, but the point I was making before was that the code had worked fine with just setting names in the sources and then suddenly stopped working. That’s why I was wondering.

aaronbell:
Something I just thought of. Previously we were setting the font family name in the source font files (eg, source.info.familyname) but I don’t believe that actually changed anything in the designspace document. Potentially, the designspace document is now being given greater priority over the source fonts info for instance generation, for whatever reason, requiring modification of the designspace naming.

DHowett:
I have been a poor steward of this repository, and I should have produced a build in the interim to make it easier for folks who want to use Cascadia.

I've been off the ball for quite a while, and it is time to get back on it. Sorry for any delay/for disappearing.

madig:
It's ok, we're all open-source here.

PR #373, opened 2020-10-01
Remove GID restrictions and auto generate CVAR

aaronbell:
This update integrates a [new VTT library vttmisc](https://github.com/aaronbell/vttmisc) with scripts to (1) overcome the GID restriction in VTT (we can now change glyph order without concern) and (2) generate a `cvar` table automatically based on the `TSIC` table. 

## PR Checklist
* [X] CLA signed.

I wrote vttmisc that addresses the GID restrictions in VTT and can generate a `cvar` table from the TSIC table. 

This update also requires several additional changes:
- the hinting source is no longer preserved as a ttx file, but is a VTT source ttf file. This allows us to reference the glyph order previously used in case that the GIDs have changed. I thought it would make sense to update this automatically on build but haven't implemented that yet so you can review how it works :). 
- I've removed the `manualhacks` function and moved the flags modification to the main function
- this (should) also includes updates from #370 
- updated requirements files
- updated features file

## Validation Steps Performed
Tested that the produced file builds properly both in vttlib and VTT itself. For CVAR generation, compared against VTT output.


madig:
Idle, tired thought: does tsic.py reimplement taking a bunch of CVTs and interpolating a CVAR? becausee varLib contains code for that already.

aaronbell:
@madig It uses the TSIC table to generate the CVAR. So yeah, kinda. I couldn't understand how varLib generated the CVAR table, so I built my own. Probably would be good at some point to figure out the varLib approach, but not at the moment. :)

madig:
I have the code lying around in a proprietary code base but basically copy-pasted the approach in varLib. I wasn't quite sure how VTT computed it, or I would have put it into vttLib. One day.

aaronbell:
I think VTT takes the same approach as varLib, FYI.

madig:
I.e. take a bunch of CVTs and their positions and compute a CVAR from just that instead of _all_ locations in a DS?

aaronbell:
I think so! Though to be fair you have to have an understanding of the entire designspace (and what is / isn’t relevant) to figure out the right CVAR entries for a single location.

DHowett:
Does this mean we have to go back to tracking the vtt code as a binary file? Is there some cool/fun way to get the best of both worlds? It was nice to be able to look at the XML sources and watch them change, but if that's too obnoxious to maintain I reckon we can  just plug our noses and go with it.

DHowett:
Aaron, I pushed a merge of master to bring in the commit that did the name stuff (just to keep history all linear and piecewise)

aaronbell:
Essentially what would need to happen is that we dump both the VTT sources and the `post` table (which is what getGlyphOrder uses). Then import both and modify the script to account for that.

Not crazy, but a bit extra work.

DHowett:
The thing I was least comfortable with was losing the vtt "source code". It feels like a step backwards . . . but I am willing to live with it assuming it's better for your workflow. I know that's a weak rejection, but I'm _slightly_ principled! 😁

aaronbell:
@DHowett Oh don't worry, I'm not backing down ;). 

Preserving the hinting data in a source file actually is much more convenient. (1) The output of VTT is a TTF source file. So to convert to a source code file I actually have to process the VTT output, and export the code. Now I could write a script to do that, but I could also just copy the VTT output TTF into the sources folder and call it a day. (2) The TTF stores the previous glyph order. Obviously I could also export that, and then reload it, but that would require a script too. 

All told, it is quite a bit simpler for me to manage it as a VTT source file. Else I have to build the font if I want to bring it into VTT to edit.

PR #380, opened 2020-10-09
Remove GDEF code from the feature files

### LINKED ISSUE #379: GDEF glyph classes are incomplete
# Environment

```none
Cascadia Code version number: 2009.22
Application (with version) used to display text: any that supports the mark feature
OS platform and version: any
Screen resolution (i.e. 220dpi):  any

Any other software?
```

# Steps to reproduce

Test the string "Ɔ̂ɔ̂ɪ̂ᴀ̂". None of the marks are placed on these base letters (and others) even though they have the correct anchors in the sources.
Test the string "m̥m̱m̲
". Theses combining marks will not be placed correctly even though the have the correct anchors in the sources.

# Expected behavior

The anchors in the sources should end up in the mark feature.

# Actual behavior

The anchors in the sources are not processed by ufo2ft when it generates the mark feature because the GDEF definition is incomplete.

base
marks

aaronbell:
This lets ufo2ft build the GDEF automatically from what is in the sources. Helpful since it means we don't have to worry about updating it constantly.

## PR Checklist
* [x] Closes #379

Resolves diacritic positioning on glyphs previously not included in GDEF
Screen Shot 2020-10-09 at 9 14 46 AM

## Validation Steps Performed
Checked in FontGoggles. Verified things are working correctly.


DHowett:
Neat! No downsides?

aaronbell:
Not that I saw.

moyogo:
Actually the GDEF definitions can also be removed from the sources/*.ufo/features.fea files.

aaronbell:
Those are autogenerated by glyphs2ufo and we override them as part of the build process. So they don't need to be removed.

PR #404, opened 2021-02-10
Help VS users get the best font rendering

AArnott:
Cascadia Code looks pretty bad in the VS editor under default settings. To help ensure developers see the font in the best rendering, this change to the README helps users realize it can look much better.

**BEFORE**

image

**AFTER**

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,8 @@ Once unzipped, right-click the font file and click "Install for all users". This
 
 For more details and app-specific instructions, [please check the wiki](https://github.com/microsoft/cascadia-code/wiki/Installing-Cascadia-Code). 
 
+**Visual Studio users:** Cascadia Code looks much better in the editor after setting this option: Tools->Options->Text Editor->Advanced->Text formatting method->Ideal.
+
 # Get involved
 Instructions on how to modify and submit an update to the Cascadia Code source is [available in the wiki](https://github.com/microsoft/cascadia-code/wiki/Modifying-Cascadia-Code).
 
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ Once unzipped, right-click the font file and click "Install for all users". This
 
 For more details and app-specific instructions, [please check the wiki](https://github.com/microsoft/cascadia-code/wiki/Installing-Cascadia-Code). 
 
-**Visual Studio users:** Cascadia Code looks much better in the editor after setting this option: Tools->Options->Text Editor->Advanced->Text formatting method->Ideal.
+**Visual Studio users:** Cascadia Code may look better in the editor after setting this option: Tools->Options->Text Editor->Advanced->Text formatting method->Ideal. This may not be necessary on high DPI screens.
 
 # Get involved
 Instructions on how to modify and submit an update to the Cascadia Code source is [available in the wiki](https://github.com/microsoft/cascadia-code/wiki/Modifying-Cascadia-Code).


DHowett:
So, this is interesting to me. Aaron went to great lengths to add hinting instructions to the font to make sure it looked good in VS, and it looks like this option _disables_ hinting. @aaronbell, what's your opinion here?

aaronbell:
Ah! One of the classic debates. 

There are two schools of thought in what constitutes "good" rendering. On one side, there is the belief that letters should have sharp edges with limited smoothing (aka, BLUR). On the other side, there is the belief that the letter shapes should be preserved as much as possible, so just to smooth / blur everything. By turning off hinting, you are switching to a blur-only approach to rendering. Microsoft has long preferred the former method (and has research to back it up showing increased legibility / readability), especially on low resolution screens, and only disables hinting at larger sizes where it no longer benefits text rendering. You'll observe that this also has a side effect of making letters bolder.

All in all, I'm glad that setting this option changes the text rendering to your preference, but I do not think it correct to tell all users to select an option that turns off hinting as it will negatively impact the experience of a vast majority of users (especially those on a <150 DPI screen rendering text at normal sizes).

@DHowett My bigger question is, why the heck does VS have a text rendering option called 'ideal' that _turns off hinting?_

aaronbell:
I recalled that I sent an email to @DHowett about Visual Studio back in Oct 2019 😄:
> Visual Studio is using an antiquated ClearType rendering mode without Y-direction antialiasing. This is the reason why text appears jaggedly in app. When I swapped fonts to Fira Code, I observed the same issues, but because Fira Code is much thinner, it is slightly less noticeable. If Visual Studio updated its system to use the DWrite default subpixel rendering and allow the GASP table within the font to instruct it how to render text, then Cascadia Code would render akin to in VSCode and Windows Terminal. I also don’t think it would significantly impact the rendering of any other font used.

Amusingly, by setting font formatting to "ideal" the poor rendering code is bypassed, and instead text is blurred, which can improve the perceived rendering for some folks. However, I still believe that this can also result in a _significantly worse_ rendering experience for others, which prevents me from recommending the 'ideal' setting. 

From my email:
> To resolve the jagginess, IMO the best fix for this situation is for Visual Studio to update their system to use the DirectWrite default subpixel text rendering and let the GASP table within the font instruct the system about how to render text. I don’t know how easy it would be for them, but that’s the right solution.

AArnott:
Interesting discussion, and TBH most went over my head. I took the screenshots and filed this PR from a standard DPI desktop screen. Now I'm looking at this issue from a HDPI Surface Book and *both* screenshots look bad, but especially the *after* version. Maybe the PNGs I pasted are being poorly enhanced for my HDPI screen. Anyway suffice it to say, "after" looked better to me on my low DPI desktop monitor.

Cascadia Code looks good in VS on this SB2 high DPI screen, without changing this setting, so evidently it perhaps only helps to switch the setting on low DPI screens.

schuelermine:
I think this would be better if you changed the text to say "can look better to some"

schuelermine:
As to suggest that the user might try it to see if they like it

AArnott:
@schuelermine Is this feedback based on the latest iteration of the PR? It *already* says it "may":

> Cascadia Code may look better in the editor after setting this option

schuelermine:
Oh OK, sorry

schuelermine:
> @schuelermine Is this feedback based on the latest iteration of the PR? It _already_ says it "may":
> 
> > Cascadia Code may look better in the editor after setting this option

@AArnott Sorry. I hadn't seen.

cntrump:
Font Family: `Cascadia Mono SemiLight` looks better

PR #463, opened 2021-05-25
Add Cascadia Italic!

### LINKED ISSUE #63: Feature Request: Provide italics version
I love the italics version of Consolas, and I use it for comments. I wish Cascadia had an italics version too.

aaronbell:
Addition of Cascadia Code Italic
cascadia-code

## PR Checklist
CASCADIA CODE ITALIC!!!
[x] Closes #63 - Italic variant added

Adds character coverage matching (and extending) the current Cascadia Code version. 

cascadia-code-italic-characters


aaronbell:
@DHowett The conflicts for the `public.background` files can be discarded. The four feature files should be deleted. And the build.py should use the incoming.

DHowett:
Alright, I'll do the resolution! Thanks!

DHowett:
I love this pluggable feature model! Clever AF

There are additionally .fea files _inside the italic font directories_ -- I suspect that those are for _the actual font software_ 😁  since it won't know about your build tricks. Those have to stay, right?

DHowett:
@aaronbell before you end up in conflict hell, you may want to (save a copy of the Hebrew/Arabic work in case Git decides it hates you, then) fast-forward your `master` branch and merge in `main` from this repo again.

The merge workflow we use on this repository destroys individual contributor histories, and is the reason your changes conflict with themselves (git can't tell that the two branches have the same changes)

DHowett:
Of course, this matters only after this PR merges :P

aaronbell:
@Dhowett

The .fea files inside the italic font directories are generated as part of the GlyphsLib conversion process. I basically ignore them entirely in favor of the hotswap feature model. You can see in the build process where I've created the list of which features are being pulled in for which font file. 

I'm not concerned about losing any Hebrew / Arabic bits and bobs namely because (1) The UFOs will be generated again in the future and will get overridden (2) All of the code is there, just commented out at current. So If you rebuild Cascadia Code, it'll include a lot of unnecessary comments (I suggest just reusing the old ship file for now).

aaronbell:
But yes, I'll see about getting my repro up to date with y'all so it isn't so confusing!

DHowett:
Ah, interesting. It does look like there are some differences in Cascadia Code when built this way... like the missing CVAR table (since it was commented out). Will that be an issue for the Italic one?

I can definitely stitch together two release packages -- the original one and the new one -- but that's not as reproduceable. I'd only do it for one release :smile:

aaronbell:
Oh! On CVAR, you need to update your local copy of vttLib. It can now generate that automatically.

I forgot I need to update requirements.txt

aaronbell:
Thanks. For some reason I can't get it to generate correctly on my end. :/

DHowett:
Oh, I understand what's happened:

https://github.com/microsoft/cascadia-code/blob/0b468d0f30de82216fd6b474796f6723de5bd585/build.py#L237-L238

We aren't merging VTT data in any longer! Only for Italic.

This looks like it may be erroneous. Should I drop that toplevel condition?

This one will still happen:

https://github.com/microsoft/cascadia-code/blob/0b468d0f30de82216fd6b474796f6723de5bd585/build.py#L240-L243

aaronbell:
Oh, shoot. Yes. I had disabled hinting for the Regular version because it was breaking (due to reasons) and would get sorted for the main Arabic release. But since the Regular hasn't been changed in this set, then the top-level condition can be removed. 

The second condition is fine—we need to differentiate between the Italic hinting source and Regular hinting source.

DHowett:
Yep, definitely. Just wanted to illustrate that I was gonna keep that all-important one. :D  Thanks!

DHowett:
So, the compiled font for Cascadia Code doesn't look terribly different from the last release; Looking it over, I can't find any comments and all the changes seem to be due to the introduction of an empty ital axis (which also makes sense).

Anyway, that's a long way to say: I'd be comfortable shipping both Code and Curve out of this build unless it's provably broken somehow!

Merging this :D

aaronbell:
Thanks Dustin! I promise the mid-June release will be much cleaner (and more complex 😂)

coxackie:
omg! when will this be released?

aaronbell:
@coxackie Just now! https://github.com/microsoft/cascadia-code/releases/tag/v2105.24

coxackie:
@aaronbell (almost) literally pissing my pants. Thank you, thank you, thank you!

adrian5:
I wouldn't know where else to post, so thanks a lot guys for your hard work in maintaining and polishing this font!

Uxorious:
That new Italic font is a big surprise - it looks like a completely different (handwritten) font.
Is that on purpose?
I'm now in the situation where the regular/bold sentences are easily readable, but anything in italics is really hard to read :-(

dmstern:
Great work, I love it! 👍

jadedarko:
I thought the new beautiful italics were from my theme, and was wondering how to use them everywhere.

Thank you for this, it's beautiful! ^_^

PR #466, opened 2021-05-25
Updating build script to remove unnecessary "Italic" from static instances

### LINKED ISSUE #465: 2105.24 - Italic Style Groups and Font File Names Incorrect
The style groups are incorrect in the new Italic fonts.
This causes the all the Italic fonts to be in their own style group rather than linked with the related Regular.
Also causes the font file names to be confused.

And the Italic typographic family is also different.

aaronbell:
Removes the unnecessary "Italic" from the static instances. 

## PR Checklist
* [x] Closes #465

For whatever reason glyphslib sets the familyName of the font to include the "Italic" name. I set the variable font instance, but not the static instances. This has been corrected, and all static instances are building as expected now.

### DIFF
--- a/build.py
+++ b/build.py
@@ -202,9 +202,13 @@ def build_font_static(
     name: str,
 ) -> None:
     prepare_fonts(designspace, name)
+
     generator = fontmake.instantiator.Instantiator.from_designspace(designspace)
     instance = generator.generate_instance(instance_descriptor)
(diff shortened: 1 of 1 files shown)


DHowett (approved):
Thanks!

mhinton:
Any chance for an updated release with this fix?

PR #491, opened 2021-06-02
Fixing regression in OT features and updating Cascadia Code Italic

### LINKED ISSUE #468: Option to use the old italic version
# Environment

```none
Cascadia Code version number: 2105.24
Application (with version) used to display text: vs code
OS platform and version: Windows 10 20H2
Screen resolution (i.e. 220dpi): 96dpi

Any other software?
```

# Steps to reproduce
The new italic font is hard to read
image

# Expected behavior
I would like to be able to use the old italic font.

# Actual behavior

### LINKED ISSUE #470: Some cursive lowercase glyphs appear to have a different slant than the others
# Environment

```none
Cascadia Code version number: 2105.24
Application (with version) used to display text: not relevant
OS platform and version: not relevant
Screen resolution (i.e. 220dpi): not relevant
```

# Steps to reproduce

1. Use Cascadia Code v2105.24
2. Write something that involves a lowercase L, lowercase Y, and possibly some others

# Expected behavior

A consistent slant for every glyph (except for block glyphs and similar "graphic" ones)

# Actual behavior

I will use this screenshot of a doc comment to illustrate:

image

In an attempt to highlight the differences in perceived slants I added vertical pipes between the lines, while forcing my editor to use a small line height.

So first of all, I want to acknowledge that this is somewhat subjective, and that I might be alone in "reading" these glyphs this way. Having said that, to me the cursive lowercase L feels almost straight. The other glyphs appear to "align" with the slanted vertical pipe characters above and below their lines (although the lowercase Y also appears a bit less slanted than the rest of the glyphs, but it's not as jarring). 

I know this is very nitpicky, and I don't want to criticize the designers of the cursive set - I think their design is cute overall, and I'm aware that font design is an art as much as a science, and more importantly that it's an incredible hassle to get it to "feel" just right. However, when quickly skimming text any cursive L's in the text really stand out a lot to me, kind of like how a single bold glyph among non-bold glyphs would stand out. They don't "disappear" into the whole word but break into individual characters, if that makes sense. And I think it's largely because of this difference in (perceived) slant.

(on a similar note the lowercase R breaks with the x-height in a way that makes `dendrogram` read a bit "wobbly", but that's probably overdoing the nitpicking)

### LINKED ISSUE #486: Regression in `Soft_Dotted` characters
### Cascadia family version

2105.24

Cascadia Code (the version with ligatures), Cascadia Mono (the version without ligatures)

### Font file format(s)

.ttf (variable), .ttf (static), .otf (static)

### Platform

macOS 10.14.6

### Other Software

_No response_

### What happened?

Characters with [the `Soft_Dotted` Unicode property](https://unicode.org/Public/13.0.0/ucd/PropList.txt) should lose their dots when followed by marks above (specifically, marks with ccc=230). Cascadia does not support this. This is a regression; see #386.

Here is <U+0069, U+0325, U+030A, U+006A, U+0325, U+030A, U+012F, U+0325, U+030A, U+0456, U+0325, U+030A, U+0458, U+0325, U+030A, U+1ECB, U+0325, U+030A>. Each soft-dotted letter is followed by two diacritics, one below and one above. All of the dots above the letters should be suppressed.
i̥̊j̥̊į̥̊і̥̊ј̥̊ị̥̊

aaronbell:

- Fixing a bug in the ccmp feature file
- Deleting the (unused) liga feature file
- Adding ccmp to the feature set in the build script.
- /f, /l, /s, /r have been updated to follow more 'generic' forms.
- previous versions have been moved to ss01 / salt OT features.
- all related glyphs have been similarly adjusted.

## PR Checklist
* [x] Closes #486
* [x] Closes #470 
* [x] Closes #468  

##Before / after

Old Italic:
Screen Shot 2021-06-03 at 2 38 29 PM

New italic forms:
Screen Shot 2021-06-03 at 2 35 13 PM


build.py
```diff
@@ -23,8 +23,7 @@
 from vttmisc import tsi1, tsic
 
 VERSION_YEAR_MONTH = 2106
-VERSION_DAY = 03
-
+VERSION_DAY = 0o3
```
DHowett, line 26: Whoa. I think we might need `3` instead (0o will blow up or act strange if we ever go above 7 (0o8 is invalid!)). Since it's not a string, the leading 0 was actually doing nothing anyway :smile:

aaronbell: Eh heh. Yeah, I was worried that if we push out two releases in a month that the "3" would end up being a 'higher version' than, say "20". Thought this might work, but yeah, nope.

aaronbell: (adjusted to `3`)

aaronbell:
Closing, as will be superseded by another PR soon. :)

PR #537, opened 2021-07-15
Correcting STAT table

### LINKED ISSUE #500: [Question] How can I use only light and light italic by default in vscode?
I have tried to build my self based on main branch, and I just allow `CascadiaMono-Roman_Light` and `CascadiaMono-Italic_Light-Italic` font in my font setting.
图片

图片

But I have issue to make use of them in vscode, all fonts in vscode is italic..
图片

If I enable `CascadiaMono-Roman` and `CascadiaMono-Italic`, this issue disappeared... But the weight is too high for me, so I just want to use a lighter version..

aaronbell:

This includes two fixes:
- [x] Closes #500 
Changes the build process to use gftools.stat instead of statmake. The issue was caused by an 'unused' italic axis in the fonts, which has been removed, but statmake gets unhappy without it there. So the process has been adjusted to a more amenable build library. 

Also:
Adjusts calt feature to ensure LIG substitutions properly function under rare conditions where the `ignore` activate incorrectly. I thought this had already been PR'd but seems not. 

For example:
`===]`
First = becomes "equal_equal.liga", and ignore would activate on second =, blocking LIG from being substituted. If the liga form appears, LIG should appear.

Turned back on the =>> =<<, etc. ligatures as there is a glyph for them now.

In other news, I fixed my git repository.


aaronbell:
Sorry. Deleted the branch :3.

PR #539, opened 2021-07-16
Fix setting to enable cursive italics in README

apcamargo:
This PR just changes the README to fix the VSCode setting to enable cursive italics.


aaronbell:
👍

Issue #704, opened 2024-01-11
Problem adding GDI+DWrite glyphs pair for contribution (help needed)


PhMajerus:
I'm trying to add some missing symbols to Cascadia to be able to submit my contribution as a pull request.
I've forked the project and managed to build it and add new characters to it just fine.

I have two main categories of symbols: Some are based on the dimensions of characters, while others are designed to fill their cell (like the U+2588 full block character).
For the ones based on characters dimensions, I looked at existing uppercase letters and found out that they should fit in the bounding rectangle 0,1420 to 1060,0.
For the ones based on a full cell, my understanding is I have to provide a GDI-based version that fits the rectangle 0,2226 to 1200,-480, as well as a DirectWrite-based version that fits the rectangle 0,1900 to 1200,-480.
The GDI one is considered the base one and should include the unicode hex value, while the DirectWrite one should have a `.stypo` suffix to its name and filename, and not include any unicode hex value.

I looked at the files referencing existing block characters in the project and believe I found all the places where I should add references to mine, including the `rclt` substututions to make it use the DirectWrite glyphs instead of the GDI ones.

Unfortunately, it appears the font I build always uses the GDI glyphs, making them the wrong height in Windows Terminal.

I would really appreciate some guidance to get this to work so I can add many symbols I have been working on.

---

I tried a simple modification to add 3 characters, U+2427, U+2428, and U+2429.

**Here is how I proceeded so far:**
(These changes can also be seen on my repository https://github.com/PhMajerus/cascadia-code)
 
### Glyph files
Added glyph files to the `glyphs` subfolder of each font (`CascadiaCode-Bold.ufo`, `CascadiaCode-BoldItalic.ufo`, `CascadiaCode-ExtraLight.ufo`, `CascadiaCode-ExtraLightItalic.ufo`, `CascadiaCode-Italic.ufo`, and `CascadiaCode-Regular.ufo`)
- `checkerB_oardD_eleteA_pple2.glif`
- `checkerB_oardD_eleteT_rs80.glif`
- `checkerB_oardD_eleteA_mstradC_pc.glif`
- `checkerB_oardD_eleteA_mstradC_pc.stypo.glif`

The Apple2 and Trs80 symbols dimensions are relative to characters, while the AmstradCpc symbol dimensions are relative to a full block. Therefore, two versions are provided to support GDI and DirectWrite-based renderers.

### Contents files
[...]

Issue #647, opened 2022-08-20
Alternate Braille glyphs optimized for pseudographics use


PhMajerus:
# Introduction and concepts

Some characters are designed especially for pseudographics, such as the old ▄ ▀ █ (as well as ▌ ▐ ░ ▒ ▓), already found in IBM PC / MS-DOS. These are often referred to as half blocks as they can be used to treat each character cell as 2 pseudo-pixels (or a 1×2 grid).

Others have been inherited from the PET / Commodore 64 and other 8-bit era computers, splitting each cell in 4 pseudo-pixels (or a 2×2 grid): ▖ ▘ ▌ ▗ ▄ ▚ ▙ ▝ ▞ ▀ ▛ ▐ ▟ ▜ █.

A higher grid density has been inherited from TRS-80 and Teletext, and now added to the Unicode Symbols for Legacy Computing, supporting 6 pseudo-pixels (or a 2×3 grid): 
![image](https://user-images.githubusercontent.com/25664275/185758814-530142aa-a176-48d9-a864-a94c612a3e27.png)
These are not yet supported, but are in other feature requests: https://github.com/microsoft/cascadia-code/issues/597 and https://github.com/microsoft/cascadia-code/issues/607

[...]

PR #705, opened 2024-01-11
Fixes for half-blocks and quadrants, and new checkerboards

### LINKED ISSUE #644: Half and quadrant block elements are not aligned on the same grid
### Cascadia family version

2111.001

Cascadia Mono (the version without ligatures)

### Font file format(s)

Windows Terminal included version (TTF (variable))

### Platform

Windows 11 build 22621.382

### Other Software

Windows Terminal 1.15.2003.0, but identical problem in the Windows Console (conhost).

### What happened?

The block elements characters are supposed to be able to mix together to be used as pseudo-graphics.
The █ character fills the whole cell, the common ▀ and ▄ characters are half blocks, and ▌and▐ are left and right halves, but the less common quadrants should be aligned on the same grid, the whole set of 16 characters (with space or nbsp) should then provide all possible combinations of the 2×2 grid:  ▖▘▌▗▄▚▙▝▞▀▛▐▟▜█.

Unfortunately, the alignment grid isn't the same for all those characters.

`"\u2580\u2584\u2596\u2597\u2598\u2599\u259a\u259b\u259c\u259d\u259e\u259f"`

image

The horizontal line splitting upper and lower halves should be identical for all those characters, basically always using the same 2×2 grid with each quadrant on or off like pixels, ideally splitting them evenly (perfectly centered).

For reference, here is the same sequence of characters using the UNSCII font (http://viznut.fi/unscii/):
image

Note some characters of this set are also shared to form the set of squot/sextant characters (2×3 grid), with U+1FB00..U+1FB3B of the Symbols for Legacy Computing. See https://github.com/microsoft/cascadia-code/issues/597 and https://github.com/microsoft/cascadia-code/issues/607.
It would make sense to rebuild them all as a single coherent set, all based on divisions of the reference █ full block character.

PhMajerus:
This is a very small update to prepare for unified mosaics, but it touches some existing glyphs.

The full block `█`, half blocks (1×2 and 2×1) `▀` `▄` `▐` `▌`, and quadrants (2×2) `▖` `▘` `▗` `▚` `▙` `▝` `▞` `▛` `▟` `▜` are designed to be aligned on the same grid. Half blocks are used to complement the quadrants to get all possible 2×2 mosaics.
As explained in #644, there was a problem with some of these not being aligned on the same grid, preventing them from getting aligned properly when combined.
In preparation of sextants (2×3) and octants (2×4), as well as separated quadrants and sextants, this update uses a single set of coordinates for the subpixels of all mosaic characters.
This is the only pull request I have planned that requires changing existing glyphs, all the future ones only add new glyphs, so I pushed this one as a small one to make it easier to review as it probably requires more attention, the next ones will contain lot of glyphs, but based on programmed patterns.

 
## References
This is the first and most difficult step towards providing the sextants as requested in #607, and octants discussed in #647.
While this update doesn't resolve either of these issues, it is the most difficult part to build a unified coordinates grid for all mosaics, which will then allow me to add the sextants and octants very quickly (they are ready, so I'm talking within days).

## PR Checklist
* [x] Closes #644
* [x] CLA signed. (I believe I did during Build 2020)
* [ ] Requires [/images/cascadia-code.png](/microsoft/cascadia-code/blob/main/images/cascadia-code.png) and/or [/images/cascadia-code-characters.png](/microsoft/cascadia-code/blob/main/images/cascadia-code-characters.png) to be updated <- Not updated as the changes are minor and wouldn't be visible, next updates will be much more interesting!
* [x] I've _mentioned_ this to @aaronbell in #704 and he seemed enthusiastic.

Many of the existing half blocks and quadrants used y=707 for their vertical center. This has been changed to y=873 for GDI versions and y=710 for DWrite versions. This is geometrically correct considering their bounding boxes are respectively [0,2226},[1200,-480] and [0,1900],[1200,-480].
These new coordinates have been tested with other mosaic characters and render correctly.

This update also adds a few characters of the Symbols for Legacy Computing, especially interesting for terminals:

- U+1FB95 Checker Board Fill (full block size)
- U+1FB96 Inverse Checker Board Fill (full block size)
- U+2427 Symbol for delete square checker board form (Apple ][ style)
- U+2428 Symbol for delete rectangular checker board form (TRS-80 style)
- U+2429 Symbol for delete medium shade form (Amstrad CPC style)

The two first ones are part of Unicode 13, while the 3 other ones have been accepted for Unicode 16 on 2022-Jan-26 (https://www.unicode.org/alloc/Pipeline.html).

## Images of the Character(s)
image
Note the existing half-blocks and quadrants are now consistent, but seem to be a bit off, with their middle y separation too high. I'm not sure if Windows Terminal handles them as special cases, or if some hinting is involved and changes the pixels snapping.
For testing, I included U+1F67F and U+1F67E to be identical to U+259A and U+259E respectively, they are literally using `<component base=.../>` to reference the existing characters, and yet they render differently, they are properly separated in 4 equal parts while the quadrants are not.
To be able to include all mosaics, we really need all of them to be rendered similarly, half blocks and quadrants are used as sextants and octants to avoid duplicating existing characters.
So I suggest we update using the new coordinates and quickly add sextants and octants, and hopefully the Terminal team can unify how they handle all characters, or simply remove any blocks-specific code they added when Atlas wasn't able to render them well yet. The new characters that Windows Terminal doesn't handle as special case seem to render really well now.

## Validation Steps Performed
Visual test on 100% and HDPI 200% displays, going through basically every zoom level using Ctrl+MouseWheel in Windows Terminal, and verified the glyphs in the built file using VTT.
I actually have sextants and octants built into a test version to verify how they all fit together and render. The pull requests are separated to make it more manageable to review the changes, but I have 386×2 other mosaic characters to submit.

Note the characters are not hinted, I wouldn't even know where to start, but they seem to behave similarly to the existing mosaic characters. They look great as it, I'm not sure whether simple geometrical shapes need hinting. Be careful though as all mosaics need to fit together, half blocks must align with quadrants, both must align with octants, half left and right must align vertically with all as well as with sextants, etc...
Whenever a new higher definition grid is introduced, they only include the missing pattern and reuse the equivalent existing patterns from existing lower resolution mosaics.


PhMajerus:
@aaronbell
I didn't succeed in refactoring the quadrants files and settled for just fixing the coordinates they use in the existing files.
The next mosaics (for next pull requests) are a lot of glyphs, so I used the same convention as the Unicode documentation, with a prefix to define which set of mosaics it is and a suffix to specify which pseudo-pixels are on, using a string of numbers where each pseudo-pixel has its own number and the sequence stays short and very easy to identify.

The sextants will be blockSextant-[digits], where digits are taken from the following grid:
```
+---+---+
| 1 | 2 |
+---+---+
| 3 | 4 |
+---+---+
| 5 | 6 |
+---+---+
```
And the octants will be blockOctant-[digits], where digits are taken from the following grid:
```
+----+----+
|  1 |  2 |
+----+----+
|  3 |  4 |
+----+----+
|  5 |  6 |
+----+----+
|  7 |  8 |
+----+----+
```

I wanted to refactor the quadrants similarly to make it easier to handle all mosaics, as follows:
`upperLeftBlock` -> `blockQuadrant-1`
`upperRightBlock` -> `blockQuadrant-2`
`lowerLeftBlock` -> `blockQuadrant-3`
`lowerRightBlock` -> `blockQuadrant-4`
`upperLeftAndLowerRightBlock` -> `blockQuadrant-14`
`upperRightAndLowerLeftBlock` -> `blockQuadrant-23`
`upperLeftAndUpperRightAndLowerLeftBlock` -> `blockQuadrant-123`
`upperLeftAndUpperRightAndLowerRightBlock` -> `blockQuadrant-124`
`upperLeftAndLowerLeftAndLowerRightBlock` -> `blockQuadrant-134`
`upperRightAndLowerLeftAndLowerRightBlock` -> `blockQuadrant-234`

But I ran into errors and couldn't find where the script referenced the older names.
While this is not required for my next submissions, I would recommand going that route when you work on the great overhaul.

PhMajerus:
Here's a test to confirm the new coordinates are right, showing sextants using the same pseudo-pixels grid:

image

Also a nice demo of what can be achieved with sextants and VT colors.

PhMajerus:
I'll include a characters picture in the next pull request, but here is an overview of what I have planned.
This pull request contains the first blue set of characters. Other groups show more quadrants, sextants, octants, separated quadrants and sextants, seven-segment digits, and large type pieces.
All of these are par of the legacy computing blocks.

cascadia-code-legacycomputing-characters

PhMajerus:
After some more days of testing with a custom build, it seems the quadrants with the old coordinates align better with the new ones I have prepared.
I don't know why using all the same coordinates doesn't work, there must be something else tweaking those glyphs.

I'd still recommend we modify them to use the same coordinates for all mosaics for better maintainability, but I'll close this pull request and start over with one that only provides new glyphs, leaving the existing ones as they are for the meantime.

PhMajerus:
After some more days of testing with a custom build, it seems the quadrants with the old coordinates align better with the new ones I have prepared.
I don't know why using all the same coordinates doesn't work, there must be something else tweaking those glyphs.

I'd still recommend we modify them to use the same coordinates for all mosaics for better maintainability, but I'll close this pull request and start over with one that only provide new glyphs, leaving the existing ones as they are for the meantime.

PR #721, opened 2024-04-09
Fix NF Mono glyph selection, NF variants STAT, PL rclt table

DHowett:
The build issue I was running into was caused by the logic that chooses when to merge Powerline glyphs into the final set before building.

It covered the following conditions:

| Conditions | Outcome                         |
| ---------- | ------------------------------- |
| Mono       | Set name only ✅                |
| Mono + PL  | Set name, merge PL ✅           |
| PL         | Set name, merge PL ✅           |
| NF         | Set name, merge PL, merge NF ✅ |

However, it missed one:

| Conditions | Outcome                         |
| ---------- | ------------------------------- |
| Mono + NF  | ❔❔❔                          | 

I've simplified the logic so that we don't make this mistake again. All non-Cascadia-Code variants will have their names overridden, and if they contain PL or NF they will get special treatment.

I've also added the NF variants to the STAT table table and fixed the generation of rclt for PL variants.


build.py
```diff
@@ -111,7 +111,7 @@ def step_set_feature_file(path: Path, name: str, instance: ufoLib2.Font) -> None
     for item in featureList:
         if "PL" in name and item == "rclt":
             featureSet += Path(path / str("rclt_PL.fea")).read_text()
-        if "NF" in name and item == "rclt":
+        elif "NF" in name and item == "rclt":
```
DHowett, line 114: This may have caused issues with the later `else`, resulting in the inclusion of `rclt.fea` after we'd already included `rclt_PL.fea` in the PL variant.

DHowett:
I also figured out why the stypo variants were being ignored in the PL/NF builds; /cc @phmajerus for the last commit

aaronbell (approved):
SGTM

aaronbell:
(Thanks so much for your investigation!!)

DHowett:
Hey, I love working on this stuff :D

PhMajerus:
Let me know if there is anything I need to do on my side when adding glyphs with stypo variants.
For now I've included the substitution of "normal" to "stypo" versions both in `features.fea::rclt` of non-italic fonts, and in the global `features/rclt.fea::rclt`. Is there any other place I need to include the stypo substitutions?

DHowett:
Just make sure it ends up in `rclt_PL.fea` as well! Otherwise the powerline/nerdfont builds won’t get the optional stypo metrics 🙂

PR #722, opened 2024-04-09
Update FONTLOG and the version number - 2404.23

DHowett:


DHowett:
You know, I might let this bake a couple more days so @PhMajerus can get more glyphs in. WDYT?

aaronbell:
I’ve no issue with that.

aaronbell (approved):

PhMajerus:
@DHowett Thanks, I'll try to merge my glyphs into a fork of the current version soon. Should be done before this weekend.

PR #723, opened 2024-04-10
New diagonals, octants, sedecimants & eights, separated quadrants & sextants, segmented digits, and large type pieces

PhMajerus:
## New diagonals, octants, sedecimants & eights, separated quadrants & sextants, segmented digits, and large type pieces (Symbols for Legacy Computing)

- Added diagonals `U+1FB3C` - `U+1FB67`
- Added octants (2×4 mosaics) `U+1CD00` - `U+1CDE5`
- Added sedecimants (4×4 mosaics) `U+1CE90` - `U+1CEAF`
- Added eights (8×8 patterns) `U+1FB70` - `U+1FB80`, `U+1FB82` - `U+1FB8B`
- Added misc. blocks `U+1FB97`, `U+1FBCE`, `U+1FBCF`
- Added separated quadrants (2×2) `U+1CC21` - `U+1CC2F`
- Added separated sextants (2×3) `U+1CE51` - `U+1CE8F`
- Added segmented digits (LED/LCD display style) `U+1CCF0` - `U+1CCF9`
- Added large type pieces (see #709) `U+1CE1A` - `U+1CE50`

This update does not modify any existing character, it is only adding
new characters from the Symbols for Legacy Computing blocks (original
and supplement).  These characters use the same unified coordinates as
my previous #708 submission. It continues the sextants with diagonal
fills that meet the sextants corners, adds octants, most of the 8×8
pixels-based lines and fills (sedecimants & eights), as well as the
separated mosaics, segmented digits, and large type pieces.  Some
existing mosaic characters are not perfectly aligned on the same grid,
and it would be best to adjust them to fit the unified grid as well, but
that is not part of this PR, and I guess we won't have the time to do
that for the next release.

Note it does not complete the original Symbols for Legacy Computing
block yet, as it does not include the extra lines/box-drawing
characters, shaded mosaics, MouseText, and some other specific symbols.
The focus has been on completing the mosaics part, including the ones
coming in Unicode 16.0, and the Large Type Pieces.

This one is quite big, containing almost all the glyphs I've been
working on at once. This is to meet the short deadline for the next
release of Cascadia Code, as discussed with @aaronbell. It contains 948
glyphs for 479 characters.

Many of the glyphs are pure geometric shapes with no artistic liberty at
all, they simply follow the unified grid and handle both GDI and DWrite
("stypo") variants.

All the glyphs have been added to the `features.fea::@NotSpace` list of
non-italic fonts, except for the segmented digits, which have been added
[...]

PR #708, opened 2024-01-16
New checkerboards, more quadrants, and sextants

PhMajerus:
## Summary of the Pull Request

This update does not modify any existing character, it is only adding new characters from the Symbols for Legacy Computing block.
This is the first pull request of a few I have ready. I unified the coordinates for sextants and octants and created the glyph files for pretty much all of the pseudo-pixels mosaic characters. The goal is to have full support for all pseudo-pixels as they are used for graphics rendering in the terminal. More goodies for terminals are also coming after the mosaics.

This contains a lot of glyphs, and the octants will be even worse. However, fear not, to avoid typos or other human error while creating hundreds of glyphs, I decided instead to create some script tools to generate them. So they should be mathematically exact, based on two bounding box sets of coordinates:
GDI versions assumes a full cell is [0,2226},[1200,-480], while DirectWrite versions assumes [0,1900],[1200,-480].

I checked all the characters in VTT, as well as in real use in Windows Terminal at 100% and 200% DPI screens and at different zoom levels. I have a set of sextants ANSI-art (or should we call those Uni-art?) files that now also display really nicely in Windows Terminal.

Here are the characters added by this pull request:
image

 
## References
Pseudo-pixels mosaic characters have been discussed in #647, and this covers all the 2×2 "quadrants" by adding some new ones, as well as all the 2×3 "sextants".
Sextants seem to be well supported on Linux, so we need these characters to ensure Windows Terminal can handle utilities using them as well.

## PR Checklist
* [x] Closes #607 
* [x] CLA signed. (I believe during Build 2020, let me know if any new signature is necessary)
* [x] Requires FONTLOG.txt to be updated
* [x] Requires [/images/cascadia-code.png](/microsoft/cascadia-code/blob/main/images/cascadia-code.png) and/or [/images/cascadia-code-characters.png](/microsoft/cascadia-code/blob/main/images/cascadia-code-characters.png) to be updated. Decided to add a separate image to avoid duplicating all the coming mosaic characters.
* [ ] I've discussed this with core contributors already. If not checked, I'm ready to accept this work might be rejected in favor of a different grand plan.

## Detailed Description of the Pull Request / Additional comments

Added 172 glyphs.

Added the new glyphs .glif files to the glyphs subfolder of each font. Being block (full-cell) symbols, they are identical in all fonts and files are just duplicated.
Block-based glyphs have both a normal (GDI) and .stypo (DWrite) variants.

[...]

PR #727, opened 2024-04-25
Adjusting half-blocks, quadrants,... coordinates to unify mosaics

### LINKED ISSUE #644: Half and quadrant block elements are not aligned on the same grid

PhMajerus:
This update adjusts the points coordinates of some previously existing blocks/mosaics characters to fit them in the same grid as used by octants and eights.
This is required because the new octants and eights from Symbols for Legacy Computing do not duplicate existing patterns and expect those existing one to join perfectly with them to provide the whole set of all possible pseudopixels mosaics.

## Description of the Pull Request

This update verifies and adjusts the existing characters that are now required to join seamlessly with the extended pseudo-pixels mosaics introduced with the symbols for legacy computing.
Some of the existing characters were not proper mosaics, most notably `▞`, where the two black rectangles overlap because they don't use the same y coordinates.
This shouldn't happen as all the mosaics are supposed to fit precisely on a unified grid.
Only a few characters required adjustments, but this PR also documents all the glyphs that have been checked to ensure alignments of all the mosaic characters.

**BLOCKS:**
`U+00A0` nbspace : already correct
`U+2588` fullBlock & fullBlock.stypo : already correct (used as reference bounding rectangle for all pseudopixels mosaics)

**HALF-BLOCKS:**
`U+2580` upperHalfBlock & upperHalfBlock.stypo : y=50% fixed from 707 to 873 (gdi) and 710 (stypo)
`U+2584` lowerHalfBlock & lowerHalfBlock.stypo : already correct (confirming upperHalfBlock y=50% was wrong)
`U+258C` leftBlock & leftBlock.stypo : already correct
`U+2590` rightBlock & rightBlock.stypo : already correct

**QUADRANTS:**
`U+2596` lowerLeftBlock & lowerLeftBlock.stypo : already correct (confirming all other corrections above and below)
`U+2597` lowerRightBlock & lowerRightBlock.stypo : y=50% fixed from 707 to 873 (gdi) and 710 (stypo)
`U+2598` upperLeftBlock & upperLeftBlock.stypo : y=50% fixed from 707 to 873 (gdi) and 710 (stypo)
`U+2599` upperLeftAndLowerLeftAndLowerRightBlock & upperLeftAndLowerLeftAndLowerRightBlock.stypo : already correct
`U+259A` upperLeftAndLowerRightBlock & upperLeftAndLowerRightBlock.stypo : y=50% fixed from 707 to 873 (gdi) and 710 (stypo)
`U+259B` upperLeftAndUpperRightAndLowerLeftBlock & upperLeftAndUpperRightAndLowerLeftBlock.stypo : y=50% fixed from 707 to 873 (gdi) and 710 (stypo)
`U+259C` upperLeftAndUpperRightAndLowerRightBlock & upperLeftAndUpperRightAndLowerRightBlock.stypo : y=50% fixed from 707 to 873 (gdi) and 710 (stypo)
`U+259D` upperRightBlock & upperRightBlock.stypo : y=50% fixed from 707 to 873 (gdi) and 710 (stypo)
`U+259E` upperRightAndLowerLeftBlock & upperRightAndLowerLeftBlock.stypo : Some y=50% fixed from 707 to 873 (gdi) and 710 (stypo), some were already correct
`U+259F` upperRightAndLowerLeftAndLowerRightBlock & upperRightAndLowerLeftAndLowerRightBlock.stypo : already correct

**OCTANTS:**
`U+2582` lowerOneQuarterBlock & lowerOneQuarterBlock.stypo : already correct
`U+2586` lowerThreeQuartersBlock & lowerThreeQuartersBlock.stypo : already correct

**EIGHTS:**
`U+2581` lowerOneEighthBlock & lowerOneEighthBlock.stypo : gdi was correct, y fixed from -183 to -182 (rounding unification for all Eights) for stypo
`U+2583` lowerThreeEighthsBlock & lowerThreeEighthsBlock.stypo : y fixed from 534 to 535 (rounding unification for all Eights) for GDI, stypo was correct
`U+2585` lowerFiveEighthsBlock & lowerFiveEighthsBlock.stypo : already correct
`U+2587` lowerSevenEighthsBlock & lowerSevenEighthsBlock.stypo : y fixed from 1887 to 1888 (rounding unification for all Eights) for GDI, stypo was correct
`U+2594` upperOneEighthBlock & upperOneEighthBlock.stypo : y fixed from 1887 to 1888 (rounding unification for all Eights) for GDI, stypo was correct
`U+1FB83` upperThreeEighthsBlock & upperThreeEighthsBlock.stypo : from my PR #723
`U+1FB86` upperSevenEighthsBlock & upperSevenEighthsBlock.stypo : from my PR #723

`U+2589` leftSevenEighthsBlock & leftSevenEighthsBlock.stypo : already correct
`U+258A` leftThreeQuartersBlock & leftThreeQuartersBlock.stypo : already correct
`U+258B` leftFiveEighthsBlock & leftFiveEighthsBlock.stypo : already correct
`U+258D` leftThreeEighthsBlock & leftThreeEighthsBlock.stypo : already correct
`U+258E` leftOneQuarterBlock & leftOneQuarterBlock.stypo : already correct
`U+258F` leftOneEighthBlock & leftOneEighthBlock.stypo : already correct

`U+2595` rightOneEighthBlock & rightOneEighthBlock.stypo : already correct
`U+1FB87` rightOneQuarterBlock & rightOneQuarterBlock.stypo : from my PR #723
`U+1FB88` rightThreeEighthsBlock & rightThreeEighthsBlock.stypo : from my PR #723
`U+1FB89` rightFiveEighthsBlock & rightFiveEighthsBlock.stypo : from my PR #723
`U+1FB8A` rightThreeQuartersBlock & rightThreeQuartersBlock.stypo : from my PR #723
`U+1FB8B` rightSevenEighthsBlock & rightSevenEighthsBlock.stypo : from my PR #723

 
## References
This fixes the inconsistent alignments problem explained in issue #644, and ensures unified grid coordinates with PR #708 and #723.

## PR Checklist
* [x] Closes #644
* [ ] Requires FONTLOG.txt to be updated (only small fixes, nothing to add to the log)
* [ ] Requires images to be updated (no new character and changes small enough that they probably wouldn't even show at that size)
* [x] I've discussed this with core contributors already. If not checked, I'm ready to accept this work might be rejected in favor of a different grand plan. Issue number where discussion took place: #644

## Validation Steps Performed
Based on purely mathematical grid coordinates already used for octants, checked visually in VTT, Terminal Preview 1.20.10822.0 and Canary 1.21.240424002-llm, and Visual Studio editor.

Note there is another related issue that impacts some of those characters, but this PR at least provides the correct glyphs and improves the situation. I believe the remaining alignment issue to be a problem in Terminal itself as it works perfectly in Visual Studio editor, and the original Cascadia Mono 2111.001 exhibits the same issue. More details about this in #644.
Having the proper coordinates at least ensures they won't induce in error someone trying to fix the Terminal rendering and expecting the alignments to work with a font using inconsistent glyphs coordinates. A believe this PR to be a step in the right direction.

### DIFF
--- a/sources/CascadiaCode-Bold.ufo/glyphs/lowerO_neE_ighthB_lock.stypo.glif
+++ b/sources/CascadiaCode-Bold.ufo/glyphs/lowerO_neE_ighthB_lock.stypo.glif
@@ -5,8 +5,8 @@
     <contour>
       <point x="0" y="-480" type="line"/>
       <point x="1200" y="-480" type="line"/>
-      <point x="1200" y="-183" type="line"/>
-      <point x="0" y="-183" type="line"/>
+      <point x="1200" y="-182" type="line"/>
+      <point x="0" y="-182" type="line"/>
     </contour>
   </outline>
 </glyph>
--- a/sources/CascadiaCode-Bold.ufo/glyphs/lowerR_ightB_lock.glif
+++ b/sources/CascadiaCode-Bold.ufo/glyphs/lowerR_ightB_lock.glif
@@ -6,8 +6,8 @@
     <contour>
       <point x="600" y="-480" type="line"/>
       <point x="1200" y="-480" type="line"/>
-      <point x="1200" y="707" type="line"/>
-      <point x="600" y="707" type="line"/>
+      <point x="1200" y="873" type="line"/>
+      <point x="600" y="873" type="line"/>
     </contour>
   </outline>
 </glyph>
--- a/sources/CascadiaCode-Bold.ufo/glyphs/lowerR_ightB_lock.stypo.glif
+++ b/sources/CascadiaCode-Bold.ufo/glyphs/lowerR_ightB_lock.stypo.glif
@@ -5,8 +5,8 @@
     <contour>
       <point x="600" y="-480" type="line"/>
       <point x="1200" y="-480" type="line"/>
-      <point x="1200" y="707" type="line"/>
-      <point x="600" y="707" type="line"/>
+      <point x="1200" y="710" type="line"/>
+      <point x="600" y="710" type="line"/>
     </contour>
   </outline>
 </glyph>
--- a/sources/CascadiaCode-Bold.ufo/glyphs/lowerS_evenE_ighthsB_lock.glif
+++ b/sources/CascadiaCode-Bold.ufo/glyphs/lowerS_evenE_ighthsB_lock.glif
@@ -6,8 +6,8 @@
     <contour>
       <point x="0" y="-480" type="line"/>
       <point x="1200" y="-480" type="line"/>
-      <point x="1200" y="1887" type="line"/>
-      <point x="0" y="1887" type="line"/>
+      <point x="1200" y="1888" type="line"/>
+      <point x="0" y="1888" type="line"/>
     </contour>
   </outline>
 </glyph>
--- a/sources/CascadiaCode-Bold.ufo/glyphs/lowerT_hreeE_ighthsB_lock.glif
+++ b/sources/CascadiaCode-Bold.ufo/glyphs/lowerT_hreeE_ighthsB_lock.glif
@@ -6,8 +6,8 @@
     <contour>
       <point x="0" y="-480" type="line"/>
       <point x="1200" y="-480" type="line"/>
-      <point x="1200" y="534" type="line"/>
-      <point x="0" y="534" type="line"/>
+      <point x="1200" y="535" type="line"/>
+      <point x="0" y="535" type="line"/>
     </contour>
   </outline>
 </glyph>
--- a/sources/CascadiaCode-Bold.ufo/glyphs/upperH_alfB_lock.glif
+++ b/sources/CascadiaCode-Bold.ufo/glyphs/upperH_alfB_lock.glif
@@ -4,8 +4,8 @@
   <unicode hex="2580"/>
   <outline>
     <contour>
-      <point x="0" y="707" type="line"/>
-      <point x="1200" y="707" type="line"/>
+      <point x="0" y="873" type="line"/>
+      <point x="1200" y="873" type="line"/>
       <point x="1200" y="2226" type="line"/>
       <point x="0" y="2226" type="line"/>
     </contour>
--- a/sources/CascadiaCode-Bold.ufo/glyphs/upperH_alfB_lock.stypo.glif
+++ b/sources/CascadiaCode-Bold.ufo/glyphs/upperH_alfB_lock.stypo.glif
@@ -3,8 +3,8 @@
   <advance width="1200"/>
   <outline>
     <contour>
-      <point x="0" y="707" type="line"/>
-      <point x="1200" y="707" type="line"/>
+      <point x="0" y="710" type="line"/>
+      <point x="1200" y="710" type="line"/>
       <point x="1200" y="1900" type="line"/>
       <point x="0" y="1900" type="line"/>
     </contour>
--- a/sources/CascadiaCode-Bold.ufo/glyphs/upperL_eftA_ndL_owerR_ightB_lock.glif
+++ b/sources/CascadiaCode-Bold.ufo/glyphs/upperL_eftA_ndL_owerR_ightB_lock.glif
@@ -4,16 +4,16 @@
   <unicode hex="259A"/>
   <outline>
     <contour>
-      <point x="0" y="707" type="line"/>
-      <point x="600" y="707" type="line"/>
+      <point x="0" y="873" type="line"/>
+      <point x="600" y="873" type="line"/>
       <point x="600" y="2226" type="line"/>
       <point x="0" y="2226" type="line"/>
     </contour>
     <contour>
       <point x="600" y="-480" type="line"/>
       <point x="1200" y="-480" type="line"/>
-      <point x="1200" y="707" type="line"/>
-      <point x="600" y="707" type="line"/>
+      <point x="1200" y="873" type="line"/>
+      <point x="600" y="873" type="line"/>
     </contour>
   </outline>
 </glyph>
--- a/sources/CascadiaCode-Bold.ufo/glyphs/upperL_eftA_ndL_owerR_ightB_lock.stypo.glif
+++ b/sources/CascadiaCode-Bold.ufo/glyphs/upperL_eftA_ndL_owerR_ightB_lock.stypo.glif
@@ -3,16 +3,16 @@
   <advance width="1200"/>
   <outline>
     <contour>
-      <point x="0" y="707" type="line"/>
-      <point x="600" y="707" type="line"/>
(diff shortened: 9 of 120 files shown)


PhMajerus:
### Visual comparison:

Before:
image

After:
image

The reason the line isn't perfectly centered is due to another issue, apparently the "normal" glyphs coordinates are used instead of the "stypo" one for those characters, even though they are substituted in `rclt`.

This can be clearly observed in Terminal Canary, which apparently doesn't crop the top extending outside of the cell anymore:
image

aaronbell (approved):
LGTM

DHowett:
Would you be able to give this a shot in Windows GVIM?

PhMajerus:
For reference, here are the coordinates I'm using for all the mosaics:

image

DHowett:
This is what I'm seeing in vim - but I need to go make sure that I have the right version installed and not a cached copy.

image

PhMajerus:
That looks suspiciously like the older coordinates, feels like a caching issue to me.

DHowett:
Actually, switching back to the old version _the block glyphs get shorter_. I wonder if the rclt/stypo switch is being compiled wrong (ala #721)

PhMajerus:
I adjusted the coordinates of both normal and stypo versions, so whichever is used, it may be taller or shorter than expected if it messes up the substitution, but it cannot have the lower and higher pseudo-pixels of `▞` at different heights.

The rclt substitution work for all my sextants, octants, etc..., and I checked some of the problematic characters and they were in the rclt of `CascadiaCode-Regular.ufo`, in the features/rclt.fea, and in features/rclt_PL.fea.

I'm testing now after just changing the code point for one of the problematic characters without changing its name, so hopefully I'll be able to see if they get substituted properly in Terminal when it's not part of the known classic block elements.

DHowett:
To your other questions in the main thread - Terminal Preview DXEngine used to scale block elements, and Atlas used to cut them off. Now DXEngine is dead, and as of 1.21 Atlas doesn't do any clipping. If you leave `builtinGlyphs` off, you'll get a 1:1 "correct" representation of the font.

PhMajerus:
I'm pretty confident Terminal is doing some special handling of classic block elements.

Here is an ansi-art using octants, and some of the characters are upper half blocks (`U+2500`).

Note how the two areas pointed in red don't align properly:
image

Now look at this version: 
image

The only difference is this:
image

And the corresponding replacement of `U+2580` to `U+E000` in the text file!
I didn't change any `rclt`, I just recompiled the font with that one codepoint change and everything else left as it.

It really looks like Terminal has leftover special handling of the classic block elements from back when tricks were used to make them fill their cell properly.

DHowett:
@lhecker You may be the expert @PhMajerus needs here

PhMajerus:
Yeah, I didn't dare to summon him, but I figured he'd be the one I need 😆
BTW, I'm available for a Team call if you want to brainstorm over this faster than through a github chat.

DHowett:
For gvim, I just installed the wrong binary out of my build output folder. GAH.

PhMajerus:
> For gvim, I just installed the wrong binary out of my build output folder. GAH.

Yeah, I didn't want to sound too confident, but that sounded like the most plausible explanation 😉

DHowett:
image

You deserve the confidence ;)

lhecker:
As it turns out this API is "at fault": https://learn.microsoft.com/en-us/windows/win32/api/dwrite_1/nf-dwrite_1-idwritetextanalyzer1-gettextcomplexity
Unfortunately, not using that API straight up doubles our CPU usage in the renderer, so I'd prefer not getting rid of it. Ideally we would find some way to mark the 0x2500 block glyphs as complex. As far as I can read dwrite's implementation, having it in gsub/gpos should mark it as complex.

DHowett:
Nah, it looks like it _is_ in gsub. We'll figure it out! Anyway, I'm merging this since it is _correct_.

Thanks all for playing!

PhMajerus:
@lhecker Thanks for looking into it.
I'm not familiar with DWrite to be of any help, but this seems to confirm that the problem isn't in the font, but instead in how DWrite handles the blocks characters that are part of the BMP.

FYI, more experiments explained in #644 seem to indicate that DWrite select the non-stypo version of these glyphs which explains why they are taller than they should. This also means that if you stretch/squash them according to `U+2588` "full block" as we discussed in https://github.com/microsoft/terminal/pull/16729, they should all render correctly as well. I'm just not sure how you would then handle the supplementary (Non-BMP) block-based characters as they cannot use the same scaling coefficients based on `U+2588` because they, on the other hand, properly use their stypo variants.

Anyway, that seems to confirm that the best is for the font to use the proper coordinates, and then find a way for AtlasEngine to find a way to handle them correctly. At least then the pseudo-pixels will align perfectly, while if we keep the previous coordinates for the base block element characters, they don't align perfectly today and won't either when AtlasEngine achieve the goal.
My coordinates used in supplementary (Non-BMP) block elements seem to perfectly align with the _Builtin Glyphs_ ones, as expected of mathematically computed coordinates, so in the meantime that's one way to get all of them aligned perfectly.

PhMajerus:
Thanks! Glad this could then be used to help test the future improvements in AtlasEngine.
Now I'm impatient to get the next Terminal Preview with the new Cascadia bundled 😁

